PR #7, opened 2016-05-22
docs: fix title example in the readme

sethvincent:
Previously the state had a property with the key `title:title` because of the model namespace. When run the example showed undefined in the placeholder on the initial page load.

The only downside of this change is that now the example doesn't demonstrate how namespaced models work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,15 +51,15 @@ productive package.
 const choo = require('choo')
 
 const app = choo()
-app.model('title', {
+app.model({
   state: {
     title: 'my-demo-app'
   },
   reducers: {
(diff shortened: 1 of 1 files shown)


sethvincent:
Oh nvm, @yoshuawuyts the pull request you just opened looks like it addresses this.

yoshuawuyts:
❤️ thanks for spotting it!

A whole bunch of people at Nodeschool yesterday took an interest in `choo` and started building stuff with it 😱 - biggest feedback I got was that namespaces were odd - so I decided to document them better and make them more rigid :sparkles: gehe, guess this was a lil wip bug

PR #8, opened 2016-05-22
Model update

### LINKED ISSUE #4: improve namespaces in models
Values in state are namespaced by `:` instead of nesting objects. Consider nesting object instead. This would mean we'd wrap `xtend` by a little if a namespace exists. Namespaces should prevent the model from touching parts outside the namespace.

This has several advantages, in terms of reasoning about the model and logging too. Also makes namespaces less icky, which is great (had to explain to several people how they currently work - turned out to be tricky; simplifying is good).

yoshuawuyts:
Major version bump. WIP
### breaking changes
- namespaces are now enforced more strictly
- models now only accept a single argument
- the `namespace` key was introduced inside of models (was prior the leading
  string in models)
- namespaced models can now only operate within themselves
- `state.location` is now namespaced under `state.app.location`

### Additions
- a whole lotta docs

---
Closes #4


yoshuawuyts:
Damn it, looks like the `"http"` example falls through - incorrect `"http"` responses must be able to set an `"error"` property somewhere to signal something went wrong - that breaks the whole idea of namespaces.

So I think what's best is if `reducers` and `state` only operate within the local namespace, but any type of `action` can be called. Bit more complex, but necessary. Should update the docs to discourage heavy cross-domain action calling though.

yoshuawuyts:
We could of course special case the `"error"` keyword, but I don't wanna mess with peoples naming schemes - don't wanna tread into frameworkism land haha

yoshuawuyts:
Welp, all examples work and I don't think there are any holes in the new namespacing code. Work is done, here. Merging!

PR #12, opened 2016-05-23
Use opts.href instead of opts.history

timwis:
Per [documentation](https://github.com/yoshuawuyts/choo#tree--appstartopts)


yoshuawuyts:
Great catch, thanks! - makes a fine point why I should get around to adding full test coverage haha :sparkles:

Issue #9, opened 2016-05-23
figure out rehydration


yoshuawuyts:
When an app is loaded from the server, figure out how to make it diff the nodes on first render, rather than overwrite the whole thing. Is probably just a call to `yo.update`.

- [x] write rehydration
- [x] update server example
- [x] update docs for server rendering to include rehydration

Extra:
- ~~write cache layer for multi routes~~

PR #13, opened 2016-05-23
toString: support rehydration

yoshuawuyts:
When an app is loaded from the server, figure out how to make it diff the nodes
on first render, rather than overwrite the whole thing. Is probably just a call
to `yo.update`.

- [x] write rehydration
- [x] update server example
- [x] update docs for server rendering to include rehydration

Extra:
- ~~write cache layer for multi routes~~

---
- Closes GH-9


yoshuawuyts:
the caching layer would probably make things too complicated hah - there's better things to do!

PR #18, opened 2016-05-25
Update form docs to use FormData

callum:
Re https://twitter.com/callumj_/status/735437016066871296 using `FormData`. Don't know how choo and send-action are about sending non-plain objects around however. Worth noting that using FormData won't allow for optimistic updates - you might have to wait for the server to respond. Currently untested, but a much cleaner and lighter approach in my view.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,20 +406,15 @@ Changing views all over the place tends to lead to messiness.
 ### forms
 Forms and lists are probably the most used concepts on any page. Together with
 links they comprise most of what can be done on web pages.
-
-Most forms contain multiple input fields, and a `submit` button that takes the
-data from the fields and submits it to the server. To collect all data from a
-form it's recommended to use the
-[get-form-data](https://github.com/insin/get-form-data) package:
 ```js
-const getFormData = require('get-form-data')
 const document = require('global/document')
 const choo = require('choo')
+const http = require('choo/http')
 const app = choo()
 
 function view (params, state, send) {
   return choo.view`
-    <form id="#login-form">
+    <form onsubmit=${onSubmit}>
       <fieldset>
         <label>username</label>
         <input type="text" name="username" autofocus>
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@callum Ohey, sincere apologies for not coming back to you directly. Thanks a lot for this PR!

I love the suggestion you make here; I definitely want to merge it - but when I also want to update the text a bit (got some ideas here, haha), and that takes a bit of time. But it'll be the next thing I do for choo, so expect this to land soon!

I merged #20 because it was a one-click improvement and people were mentioning forms on twitter. Thanks again for this PR - I really like it! :sparkles:

yoshuawuyts:
Hmmm, if you know the schema then the following is possible:
```js
  function onSubmit (e) {
    e.preventDefault()
    const data = new window.FormData(e.target)
    console.log(data.get('password'))
    send('login', { data: data })
  }
```
Can also iterate over the value of `data.entries()` using:
```js
const entries = data.entries()
const res = {}
while (true) {
  var res = entries.next()
  if (res.done) break
  res[res.value[0]] = value[1]
}
```
I agree with the notion that non-objects shouldn't be passed around if possible; exceptions being stuff like files which for obvious reasons cannot be cast to primitives.

I think we should mention all of this stuff, as it appears forms are madness to deal with - can't imagine how people that get started with JS and programming in general make any sense of the browser APIs.

yoshuawuyts:
Merged in https://github.com/yoshuawuyts/choo/commit/bd28e93cd19335d3ca69f4cbda884e6faaa160ea :sparkles:

yoshuawuyts:
ps congrats on getting commit #100 in :D

callum:
Perfect, thanks. The `FormData` API isn't great, I agree. If you want to get a specific field's value, you can quite simply do:

```js
const pass = e.target.elements.password.value
```

...providing that input has the name 'password'

callum:
`FormData` is just a much simpler way to wrap it all up and send it to the server as if it were being sent as normal, without AJAX.

yoshuawuyts:
@callum ahhh yeah I like that too!

PR #20, opened 2016-05-26
Use submit hook instead of click hook in form example

timwis:
Your call, but I think this approach is a little better, and catches "enter" presses (rather than just clicks)

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ const app = choo()
 
 function view (params, state, send) {
   return choo.view`
-    <form id="#login-form">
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Yeah, this looks great! - Thanks a lot! :sparkles:

timwis:
:O wow, sry @callum, didn't imagine someone else would have submitted a PR for this! I had no idea about `FormData` though so thanks for the pointer.

callum:
No problem at all @timwis. Glad you learned something!

Issue #1, opened 2016-05-13
lifecycle hooks


yoshuawuyts:
Every once in a while we need hooks to manage things like animations:
https://github.com/maxogden/yo-yo/issues/18

yoshuawuyts:
For managing complex animations people should probably use a finite state machine. For reaction to DOM nodes being removed (e.g. trigger an event to load / unload data) use [MutationObservers](https://developer.mozilla.org/en/docs/Web/API/MutationObserver) ([mdn](http://caniuse.com/#feat=mutationobserver).

## See Also
- [f1](https://github.com/jam3/f1)
- [fsm-event](https://github.com/yoshuawuyts/fsm-event)

yoshuawuyts:
https://github.com/shama/on-load

timwis:
What about firing an ajax request when a view is shown? Can't just put it in the view logic because then it fires again every time the state changes, and you end up with an endless loop if that ajax changes the state. In react, you'd use [`componentDidMount`](https://facebook.github.io/react/docs/component-specs.html#mounting-componentdidmount). What do you recommend with choo?

yoshuawuyts:
>What about firing an ajax request when a view is shown?

Untested, but I reckon [on-load](https://github.com/shama/on-load) does the job correctly. Given that unless the component has mutated it should run just fine. I'd hook a `send()` call to the event. If for some reason this turns out to be a bit unreliable, making the action itself idempotent is recommended so that even if the action fires nothing bad happens.

> even when there is no "old tree"

What do you mean? There should always be a tree - or is this a conflict between the first call to fetch data and the tree being rendered? If this is a bug I'd definitely want to fix this - could you clarify? Thanks!

timwis:
> making the action itself idempotent is recommended so that even if the action fires nothing bad happens.

In my case, the action is "fetch a list of tables," which is idempotent. The problem is that once they're fetched, the state changes, so the view re-renders, which triggers the fetch, the state change, the re-render, and an endless loop :P

> I reckon on-load does the job correctly

Interesting -- do you mean something like this?

```javascript
module.exports = (params, state, send) => {
  const tree = choo.view`<div>Foo</div>`
  onload(tree, () => send('tables:fetch'))
  return tree
}
```
If so, I may be missing something, because it doesn't appear that changing views triggers the `onload` library.

> What do you mean? There should always be a tree

I tried to get around the "endless loop" using this approach:

```javascript
module.exports = (params, state, send) => {
[...]

PR #26, opened 2016-05-28
Don't update a tree if it isn't in the DOM yet

timwis:
This can occur if the state changes before the first view is rendered per #1, in which case it throws the error `Cannot read property 'nodeType' of null`

### DIFF
--- a/index.js
+++ b/index.js
@@ -158,7 +158,7 @@ function choo () {
       const oldTree = document.querySelector('#' + rootId)
       const newTree = _router(state.app.location, state, send)
       newTree.setAttribute('id', rootId)
-      yo.update(oldTree, newTree)
+      if (oldTree) yo.update(oldTree, newTree)
     }
   }
 


yoshuawuyts:
This is one way of solving it, buttttt.... is it __the__ way of solving it? Perhaps an `assert()` with a descriptive message would work just as well? It feels like side effects are randomly being triggered here, which would mean an `if()` statement would be a band-aid solution, rather than alert of actual inconsistencies.

So something that's been bothering me is how to initialize the whole update loop. Ideally once the application has done loading all the things, the first update loop is called to fetch data from the API and what not. Because `send()` is only available within `choo`, the logical place to trigger this would be from a subscription:
```js
app.model({
  subscriptions: [
    function (send) {
      document.addEventListener('DOMContentLoaded', function (e) {
        console.log(e)
      })
    }
  ]
})
```
Does this make sense? Is what I'm saying true for your use case?

timwis:
Not quite sure I understand. I think the problem here is that state changes can occur before the first render, but the "on state change callback" is triggered regardless and runs `yo.update()` even though there is no tree to update.

yoshuawuyts:
Yup, your analysis is correct. Our solutions are different though.

So this patch introduces possible data races. If for some reason data a state mutation takes place before the DOM tree is appened, the new changes won't be displayed until the next render. Instead the changes should only be kicked off after the DOM is ready to be mutated. It's a mild difference, but makes for a clearer mental model. I'm going to close this patch for now, and write one which documents this behavior and provides a meaningful error.

Huge thanks though; walking through all of the erroneous behavior has been extremely helpful. Cheers! :sparkles:

timwis:
Glad to be helpful - looking forward to it, thanks

timwis:
Hey @yoshuawuyts, I was just mentally planning an app I'm thinking of building in choo (for live election results), and ended up back at this question. I'm hoping you can point me in the right direction.

So I'm planning to build an app that uses firebase as a backend. According to choo's docs, a `subscription` would be the best place to (a) initialize firebase and (b) fetch/watch data.
```javascript
const Firebase = require('firebase')

app.model({
  subscriptions: [
    (send) => {
      const firebase = new Firebase('https://...')
      firebase.on('value', (snapshot) => send('value', snapshot.val()))
    }
  ]
})
```
The problem is that there's a race condition: if somehow my request to firebase resolves before the initial render, it will trigger the error described above.

Per #30, do you suggest that I wrap my subscription in a `DOMContentLoaded` listener, like this?
```javascript
const Firebase = require('firebase')

app.model({
  subscriptions: [
    (send) => {
      document.addEventListener('DOMContentLoaded', (e) => {
        const firebase = new Firebase('https://...')
        firebase.on('value', (snapshot) => send('value', snapshot.val()))
      })
    }
  ]
})
```
If so, wouldn't that apply to the subscription examples in the readme, like [websockets](https://github.com/yoshuawuyts/choo#websockets), SSE, etc.?

yoshuawuyts:
Hah, dammit - you're right. Perhaps all subscriptions in `choo` should only be kicked off once the DOM is loaded making _all_ of this a non-issue haha. PR welcome!

timwis:
Might I go a step further and suggest that it's not the subscriptions that are causing the problem (nothing wrong with fetching data before the DOM is ready), nor the updates to the state, but the fact that the `onchange` handler is trying to update the DOM before it's rendered? If so, it might make sense to bind `onchange` inside the `DOMContentLoaded` hook instead.

yoshuawuyts:
Ah yeah, that's fair - but we must definitely ensure that no more than one call to `render` is passed when the listener kicks off. Bit more complicated I reckon - not sure if using a `raf` call (such as [virtual-raf](https://github.com/yoshuawuyts/virtual-raf)) would help with that; PRs for this welcome - think I'm gonna close #32 for now then.

PR #27, opened 2016-05-29
router: suppress re-render on same state

### LINKED ISSUE #25: Effects shouldn't trigger re-render
According to the [documentation on effects](https://github.com/yoshuawuyts/choo#effects), `Unlike reducers they cannot modify the state`. However, because [send-action](https://github.com/sethvincent/send-action/blob/master/index.js) calls the `onchange` handler every time `send()` is called, choo [re-renders the view](https://github.com/yoshuawuyts/choo/blob/master/index.js#L157-L162), which is of course unnecessary since the state hasn't changed.

[Here is a demo](http://requirebin.com/?gist=a6267d63da432ed2be0eaf2a6ef65f20) that shows the view is re-rendered by the `get` effect. A second later, it gets re-rendered by the `receive` effect, where the state actually changed.

One solution here may be to post an issue in the send-action repo that `onchange` shouldn't be called unless the state's actually changed. But another way to think of it is that choo's effects should not trigger the `onchange` handler at all. Thoughts?

yoshuawuyts:
- views will no longer be re-rendered if the state hasn't changed
- `oldState` is passed as the final argument in a view. E.g. `params,
  state, send, oldState`. This is mostly because it will usually only be
  used for optimizing - changing the order would be a breaking change :(

Closes #25 

Minor version bump

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ const sheetRouter = require('sheet-router')
 const document = require('global/document')
 const href = require('sheet-router/href')
 const sendAction = require('send-action')
+const mutate = require('xtend/mutable')
 const assert = require('assert')
 const xtend = require('xtend')
 const yo = require('yo-yo')
@@ -118,6 +119,7 @@ function choo () {
     function handleAction (action, state, send) {
       var reducersCalled = false
       var effectsCalled = false
+      const newState = xtend(state)
 
       // validate if a namespace exists. Namespaces
       // are delimited by the first ':'. Perhaps
@@ -132,9 +134,12 @@ function choo () {
       const _reducers = ns ? reducers[ns] : reducers
       if (_reducers && _reducers[action.type]) {
         if (ns) {
-          const newState = _reducers[action.type](action, state[ns])
-          state[ns] = xtend(state[ns], newState)
-        } else state = xtend(state, reducers[action.type](action, state))
+          const reducedState = _reducers[action.type](action, state[ns])
+          if (!newState[ns]) newState[ns] = {}
+          mutate(newState[ns], xtend(state[ns], reducedState))
+        } else {
+          mutate(newState, reducers[action.type](action, state))
+        }
         reducersCalled = true
       }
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -132,9 +134,12 @@ function choo () {
       const _reducers = ns ? reducers[ns] : reducers
       if (_reducers && _reducers[action.type]) {
         if (ns) {
-          const newState = _reducers[action.type](action, state[ns])
-          state[ns] = xtend(state[ns], newState)
-        } else state = xtend(state, reducers[action.type](action, state))
+          const redState = _reducers[action.type](action, state[ns])
```
timwis: Took me a minute to figure out what you meant by `redState` - perhaps a name like `reducedState` may improve readability?

timwis:
@yoshuawuyts thanks for jumping on this! I can't quite figure out what the new `mutate` logic is doing differently. My first thought was that, rather than cloning the state and altering it (the immutable way), you were going to operate on the original state object so that you could check to see if it's the same object reference (the `===` operator) but that doesn't appear to be what's happening since you clone `state` into `newState` before-hand.

```javascript
const model = {
  state: {
    value: 0
  },
  reducers: {
    increase: (action, state) => ({value: state += action})
  }
}

send('increase', 5) // should trigger re-render
send('increase', 0) // should not trigger re-render
```
If I'm reading this pull request's changes correctly, both `send`s would trigger re-render, no?

yoshuawuyts:
> If I'm reading this pull request's changes correctly, both sends would trigger re-render, no?

Yup, because a reducer was triggered, in most cases a value in the state would change. But now that the `oldState` is also passed as a tailing argument, it means the path has been opened to optimize further down the tree:
```js
var _view = null
function createView (params, state, send, oldState) {
  if (_view && state.someValue === oldState.someValue) return _view
  _view = choo.view`<div>something ${state.someValue}</div>`
  return _view
}
```
To make it behave consistently lifecycle hooks should probably introduced too so that it can be dumped from memory when it's unmounted; but that's a different issue (#1).

Also not sure about the order of arguments here; also feel like `params` could be appended to state, soooo.... getting to a possible api of `view = (state, oldState, send) => choo.view` is a different concern hah.

timwis:
Interesting - not sure what the changes on L137-142 are for if that's the case, but perhaps those are just unrelated optimizations.

For what it's worth, I would expect it to re-render _only_ if the state _actually_ changed (assuming you can also imagine scenarios where a reducer doesn't actually change the state - say, because the user pressed the "calculate" button without changing the value). I imagine the only way to ensure that is to do a deep comparison of the state object, which I understand is expensive. But I wonder what's more expensive: a deep comparison or the DOM tree comparison that occurs when you run `yo.update()`.

(Hope this is a helpful discussion)

index.js
yoshuawuyts: Fair enough

yoshuawuyts:
> But I wonder what's more expensive: a deep comparison or the DOM tree comparison that occurs when you run yo.update().

This is all discussing optimizations for an edge case: the chance that you call a function that modifies state that doesn't end up modifying the state is slim.

The new mutate logic makes sure the old state is persisted as-is, and new changes applied to a fresh object which is then conditionally returned. 

I feel like for now this is the best we can do - gonna merge this :sparkles:

timwis:
Sounds good - sorry to belabor the point. Thanks for addressing this!

PR #30, opened 2016-05-31
start: throw err for render races

yoshuawuyts:
- checks if the oldTree exists and throws a human readable error if not the case
- add docs on how to handle this situation
- check is disabled through `unassertify` for production builds

Thanks!

---
- supersedes https://github.com/yoshuawuyts/choo/pull/26#issuecomment-222692979
- closes #26


yoshuawuyts:
Released as `v2.1.4` :sparkles:

Issue #32, opened 2016-06-01
load subscriptions on document ready


yoshuawuyts:
Per https://github.com/yoshuawuyts/choo/pull/26#issuecomment-222854723, all `subscriptions` should only be started once the DOM has loaded:
```js
document.addEventListener('DOMContentLoaded', (e) => {})
```
Saves a bunch of needless boilerplate on the consumer's side.

yoshuawuyts:
Closing as per https://github.com/yoshuawuyts/choo/pull/26#issuecomment-222957154

yoshuawuyts:
ehhh, no no actually this issue should stay open. From the linked issue:
> Ah yeah, that's fair - but we must definitely ensure that no more than one call to render is passed when the listener kicks off. Bit more complicated I reckon - not sure if using a raf call (such as virtual-raf) would help with that; PRs for this welcome - think I'm gonna close #32 for now then.

yoshuawuyts:
Oops, merged this but never released the minor version. https://github.com/yoshuawuyts/choo/pull/55

yoshuawuyts:
Released in `v2.3.0`

PR #33, opened 2016-06-01
subscriptions: wait for DOM loaded

yoshuawuyts:
Closes GH-32

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,7 +81,9 @@ function choo () {
       if (model.subscriptions) {
         assert.ok(Array.isArray(model.subscriptions, 'subs must be an array'))
         model.subscriptions.forEach(function (sub) {
-          sub(send)
+          document.addEventListener('DOMContentLoaded', function () {
+            sub(send)
+          })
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -81,7 +81,9 @@ function choo () {
       if (model.subscriptions) {
         assert.ok(Array.isArray(model.subscriptions, 'subs must be an array'))
         model.subscriptions.forEach(function (sub) {
-          sub(send)
+          document.addEventListener('DOMContentLoaded', function () {
```
timwis: I wonder if swapping these two lines would work?
```javascript
document.addEventListener('DOMContentLoaded', function () {
  model.subscriptions.forEach(function (sub) { sub(send) }
}
```
That way you avoid creating an event listener for every subscription and hopefully achieve the same result.

yoshuawuyts: Oh yeah, that's a fair point :sparkles:

yoshuawuyts:
Closed as per #32

PR #38, opened 2016-06-04
Fix #37 `rootId` should clear if only `opts` is passed

### LINKED ISSUE #37: First argument of start() is not optional
The [docs](https://github.com/yoshuawuyts/choo/#links) suggest you can pass options as the first parameter to `start`, ie. `app.start({ href: false })`, but doing so throws the following errors:

`Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'`
`Uncaught TypeError: rootId.replace is not a function` 

It appears to be because [this line](https://github.com/yoshuawuyts/choo/blob/master/index.js#L55) begins to make the first argument optional, but [this line](https://github.com/yoshuawuyts/choo/blob/master/index.js#L98) uses the second arg as if it were the first.

Happy to submit a PR to fix this, but can't do it right this second, so just wanted to post the issue in the meantime.

emilbayes:
`rootId` is optional to `start`, but has lower precedence than `opts`. Currently it is not cleared when the value is passed to `opts`

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,7 +52,10 @@ function choo () {
   // start the application
   // (str?, obj?) -> DOMNode
   function start (rootId, opts) {
-    if (!opts) opts = rootId
+    if (!opts) {
+      opts = rootId
+      rootId = null
+    }
     opts = opts || {}
     const name = opts.name || 'choo'
     const initialState = {}


yoshuawuyts:
Brilliant, thanks!

yoshuawuyts:
Published as `v2.1.5` :sparkles:

timwis:
beat me to it 😛 thanks @emilbayes

Issue #21, opened 2016-05-26
Programmatic navigation doesn't change URL


timwis:
I have a form that, when submitted, saves the contents to the state (via `send`) and redirects you to another page. The documentation on [router] explains how to programmatically navigate the page:

> If you want to modify the location programmatically the reducer for the location can be called using `send('app:location', { location: href })`.

Doing this doesn't change the URL, however, so if the user reloads or bookmarks the page, they won't be taken to the right place. Would you recommend adding a line such as:

`window.history.pushState(null, null, '/foo')`

Or is there a way to accomplish that with a single line?

yoshuawuyts:
Yeah, I think this is good. It creates a slight duplication of code with https://github.com/yoshuawuyts/sheet-router/blob/master/href.js#L27 but I think that's acceptable for the sake of modularity.

Reckon something like:
```js
const newLocation = action.location.replace(/#.*/, '')
if (window && (window.location !== newLocation) {
  window.history.pushState({}, null, href)
}
```

on https://github.com/yoshuawuyts/choo/blob/master/index.js#L189 should do the trick - you reckon?

timwis:
That sounds good to me, though I wonder if it would be worth adding a `navigate()` method to the router itself so that you can abstract the logic for how navigation works (rather than accessing `history` in choo). That would certainly make [support for hash history](https://github.com/yoshuawuyts/sheet-router/issues/16) easier.

yoshuawuyts:
I think for the sake of modularity it's not wise to add any knowledge of `window.location` to the router itself, but let it live in a method outside of it that can instrument the router as needed.

Back when `wayfarer` was still a lil cub project, [hash-match](https://github.com/sethvincent/hash-match) was created - perhaps it'll come in useful here.

yoshuawuyts:
We talked over this in IRC; and making a two-way binding to the history API from a reducer is maybe not the best. It might be easier to just recommend using the DOM API. We can always revisit this in the future, but I suggest we close this for now (and document it somewhere). Cheers!

PR #39, opened 2016-06-04
location: set browser location

yoshuawuyts:
Set the browser location when called programmatically.

- Closes GH-21

@timwis could you verify this solves your issue? Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ const sheetRouter = require('sheet-router')
 const document = require('global/document')
 const href = require('sheet-router/href')
 const sendAction = require('send-action')
+const window = require('global/window')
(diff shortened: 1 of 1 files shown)


timwis:
Hey @yoshuawuyts, this does solve the issue that I posted, however it seems to break the "back" and "forward" button functionality, probably because it's conflicting with [href.js](https://github.com/yoshuawuyts/sheet-router/blob/master/href.js) and pushing the state twice.

timwis:
Note that this behaviour could/should be impacted by #40

yoshuawuyts:
Closing as per https://github.com/yoshuawuyts/choo/issues/21#issuecomment-228529976

PR #40, opened 2016-06-04
Hash routing

timwis:
This pull request leverages the recent work of @nlocnila in yoshuawuyts/sheet-router#18 to add hash routing support. The current code works, but could be optimized. Wanted to get feedback on the direction first.

Note that to use this, users must use `app.start(null, {history: false, href: false})`. Otherwise `href` hijacks the anchor tags.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,8 @@ const history = require('sheet-router/history')
 const sheetRouter = require('sheet-router')
 const document = require('global/document')
 const href = require('sheet-router/href')
+const hash = require('sheet-router/hash')
+const hashMatch = require('hash-match')
 const sendAction = require('send-action')
 const mutate = require('xtend/mutable')
 const assert = require('assert')
@@ -205,6 +207,13 @@ function appInit (opts) {
   // enable HTML5 history API
   if (opts.href !== false) pushLocationSub(href)
   if (opts.history !== false) pushLocationSub(history)
+  if (opts.hash !== false) {
+    pushLocationSub(function (navigate) {
+      hash(function (fragment) {
+        navigate(hashMatch(fragment))
+      })
+    })
+  }
 
   return model
 
--- a/index.js
+++ b/index.js
@@ -189,9 +189,16 @@ function choo () {
 // initial application state model
 // obj -> obj
 function appInit (opts) {
+  var initialLocation
+  if (opts.history !== false) {
+    initialLocation = document.location.href
+  } else {
+    initialLocation = hashMatch(document.location.hash)
+  }
+
   const model = {
     namespace: 'app',
-    state: { location: document.location.href },
+    state: { location: initialLocation },
     subscriptions: [],
     reducers: {
       // handle href links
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -204,7 +213,16 @@ function appInit (opts) {
   // enable catching <href a=""></href> links
   // enable HTML5 history API
   if (opts.href !== false) pushLocationSub(href)
-  if (opts.history !== false) pushLocationSub(history)
+  if (opts.history !== false) {
+    pushLocationSub(history)
+  // otherwise enable hash history
+  } else {
+    pushLocationSub(function (navigate) {
```
timwis: The only way I can think to shorten this like the other two is to use function composition, which would require a `compose` function or importing one from a library.

index.js
```diff
@@ -187,9 +189,16 @@ function choo () {
 // initial application state model
 // obj -> obj
 function appInit (opts) {
+  var initialLocation
+  if (opts.history !== false) {
```
timwis: Some redundancy here as `opts.history !== false` is performed below too.

yoshuawuyts: perhaps a ternary if it fits?

index.js
```diff
@@ -204,7 +213,16 @@ function appInit (opts) {
   // enable catching <href a=""></href> links
   // enable HTML5 history API
   if (opts.href !== false) pushLocationSub(href)
-  if (opts.history !== false) pushLocationSub(history)
+  if (opts.history !== false) {
+    pushLocationSub(history)
+  // otherwise enable hash history
+  } else {
```
yoshuawuyts: perhaps have `opts.hash === true` as the trigger? can imagine there's situations where neither is desirable to enable

index.js
yoshuawuyts: nah, it's probably ok (':

yoshuawuyts:
yeah, if this works then sure!

timwis:
Updated per your feedback. If there aren't any other implications I'm overlooking, I can add this to the docs and the PR will be ready to merge.

index.js
```diff
@@ -190,9 +192,13 @@ function choo () {
...
+    ? hashMatch(document.location.hash)
+    : document.location.href
+
   const model = {
     namespace: 'app',
-    state: { location: document.location.href },
+    state: { location: initialLocation },
     subscriptions: [],
     reducers: {
       // handle href links
```
timwis: Not sure the `.replace(/#.*/, '')` is necessary anymore with hash-match in use, though I'm not sure what purpose this served originally if it was meant to be used with browser history.

yoshuawuyts: Oh yeah, that was to support `<a href="#"></a>` style links; not sure it's needed anymore, but eh - let's not remove it quite yet :sparkles:

yoshuawuyts:
Oh damn, last thing for this: we need to update the docs

timwis:
Doc update added 📜

yoshuawuyts:
Merged in `v2.2.0` - thanks a lot @timwis :sparkles:

PR #43, opened 2016-06-13
Fix order of arguments to effects

toddself:
This confused me a bit since it was the reverse of the `action` and `state` arguments for reducers.


yoshuawuyts:
Haha - oops. Thanks for fixing! 👀 :sparkles:

PR #46, opened 2016-06-15
Make state immutable in reducers/effects

toddself:
Call Object.freeze on state when passing it into to the reducers or effects
so that state cannot be modified except when being returned from an emitter.

Add tests to prove state is immutable in these scenarios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,10 @@ and has access to the full application state. Try and keep the logic in these
 bulk of your logic will safely shielded, with only a few points touching every
 part of your application.
 
+Within the scope of `reducers` and `effects`, you can access the current `state`,
+however, the object is frozen, disallowing changes. To update the state, you 
+must return an object with your updates from a `reducer`.
+
 ## Effects
 Side effects are done through `effects` declared in `app.model()`. Unlike
 `reducers` they cannot modify the state by returning objects, but get a
--- a/index.js
+++ b/index.js
@@ -139,19 +139,19 @@ function choo () {
       const _reducers = ns ? reducers[ns] : reducers
       if (_reducers && _reducers[action.type]) {
         if (ns) {
-          const reducedState = _reducers[action.type](action, state[ns])
+          const reducedState = _reducers[action.type](action, Object.freeze(state[ns]))
           if (!newState[ns]) newState[ns] = {}
           mutate(newState[ns], xtend(state[ns], reducedState))
         } else {
-          mutate(newState, reducers[action.type](action, state))
+          mutate(newState, reducers[action.type](action, Object.freeze(state)))
         }
         reducersCalled = true
       }
 
       const _effects = ns ? effects[ns] : effects
       if (_effects && _effects[action.type]) {
-        if (ns) _effects[action.type](action, state[ns], send)
-        else _effects[action.type](action, state, send)
+        if (ns) _effects[action.type](action, Object.freeze(state[ns]), send)
+        else _effects[action.type](action, Object.freeze(state), send)
         effectsCalled = true
       }
 
--- a/tests/z-disabled-browser.js
+++ b/tests/z-disabled-browser.js
@@ -1,0 +1,55 @@
+const tape = require('tape')
+const choo = require('../')
+
+tape('should render on the client', {skip: true}, function (t) {
+  t.test('state should not be mutable', function (t) {
+    t.plan(3)
+
+    const app = choo()
+    const state = {
+      foo: 'baz',
+      beep: 'boop'
+    }
+    app.model({
+      state: state,
+      reducers: {
+        mutate: (action, state) => {
+          state.foo = 'zap'
+          return {}
+        },
+        noMutate: (action, state) => {
+          return {beep: 'poob'}
+        }
+      },
+      effects: {
+        effectMutate: (action, state) => {
+          state.foo = 'zap'
+        }
+      }
+    })
+
+    app.router((route) => [
+      route('/', function (params, state, send) {
+        const okd = (evt) => {
+          if (evt.key === 'a') {
+            send('mutate')
+          } else {
+            send('noMutate')
+          }
+        }
+        return choo.view`<span class="test" onkeydown=${okd}>${state.foo}:${state.beep}</span>`
+      })
+    ])
+    document.body.appendChild(app.start())
+    const $el = document.querySelector('.test')
+    const mutate = new window.KeyboardEvent('keydown', {key: 'a'})
+    const noMutate = new window.KeyboardEvent('keydown', {key: 'b'})
+    const effectMutate = new window.KeyboardEvent('keydown', {key: 'c'})
+    $el.dispatchEvent(mutate)
+    t.equal($el.innerText, 'baz:boop', 'state did not mutate')
+    $el.dispatchEvent(noMutate)
+    t.equal($el.innerText, 'baz:poob', 'state was updated from a return')
+    $el.dispatchEvent(effectMutate)
+    t.equal($el.innerText, 'baz:poob', 'state was not updated from an effect')
+  })
+})


toddself:
So! The current tests run in node which means you can't call `send`.

So I made a test file that you can browserify and run in the client.

However, setting up automated testing for browser tests is _super_ opinionated so I didn't set up a way to actually, you know, *run* the tests.

We use [karma](https://karma-runner.github.io/0.13/index.html) for scripto for client testing (and xvfb for doing it headless on CI).  However, karma is _very_ heavy and requires a TON of modules and plugins.  but it also is extremely reliable, and gets you coverage reports easily.  But, since it's also XVFB for automated testing, I still can't get it set up correctly on travis-ci (but can do it on circle-ci).

I've also done testling, which is super minimal, but doesn't get you coverage easily and still requires XVFB.

Zuul is great, but you can't run it in an automated fashion without setting up sauce labs support (I have an extremely old branch where I tried to implement local test running but never finished).

So like, um, I have no clue how you want to commit to running these tests, but like this will open it and run it...

```
echo '<title>test</title><body><script src="bundle.js"></script>' > index.html
../node_modules/bin/browserify browser.js > bundle.js
open index.html
```

(Or xdg-open if you're on linux. I have no clue how to do this on windows lol)

yoshuawuyts:
Ooh, love this patch - minor version I reckon. Just needs tests fixed and we can merge.

Browser tests are def tricky - perhaps just add these tests for now, but exclude them from CI? - that way Travis stays green and we can save ourselves some work in the future.

I've done a bit of work for tests / coverage here - https://github.com/yoshuawuyts/choo/pull/24. It uses `smokestack` which can integrate with saucelabs; I reckon using `sauce` might be the easiest?

queckezz:
Another alternative to consider might be [tape-run](https://github.com/juliangruber/tape-run). Basically runs your code through a headless electron instance. No integration with saucelabs ofc though, so v8 only. I run it like this:

```
browserify --debug test/index.js | tape-run --render="tap-spec"
```

toddself:
Oooooh I hadn't seen tape-run and it's instructions for using electron! That would greatly simplify some other only-chrome stuff in my life :).

Interestingly, you can't easily mark a test as "skip" when you run tape tests through the node executable. Adding `{skip: true}` at either level makes node simply exit after running browser test (since it comes first in the glob), and even wrapping it in an `if` statement causes node to exit after processing the file.

I renamed it so it comes later in the glob pattern (and since node will exit after the first file it will never execute), but this feels really hacky?  

Better ideas or solutions?

Also added some documentation to explain that `state` is frozen within a `reducer` or an `effect` just to be explicit about it.

toddself:
So I tried to use Zuul since then it can run in sauce labs for free using "open sauce" which means you can target a ton of browsers, but Zuul uses an older version of browserify-istanbul, which relies on an older version of esprima which is unable to parse template tag syntax.

I opened a PR to update to the latest version of browserify-istanbul, but that's a dead-end until then defunctzombie/zuul#289

I do have this running under karma in my project, but the list of requirements is pretty heavy (although some sub-deps of choo require karma already so we're installing it as it is).

Here's the minimal list I've found I need to just support tape and istanbul via browserify:
```
    "browserify": "^13.0.1",
    "browserify-istanbul": "^2.0.0",
    "istanbul": "^0.4.3",
    "karma": "^0.13.22",
    "karma-browserify": "^5.0.5",
    "karma-chrome-launcher": "^1.0.1",
    "karma-coverage": "^1.0.0",
    "karma-script-launcher": "^1.0.0",
    "karma-tap": "^1.0.4",
    "tape": "^4.5.1",
```

That feels kind of heavy, but it works.  Except for in travis-ci.

yoshuawuyts:
This patch is looking great by the way - testing, merging and publishing as minor :sparkles:

yoshuawuyts:
Ah yeah, so the problem here is that `Object.freeze` mutates an existing object, rather than create a new copy - the fact that a single action can trigger both an `effect` and `reducer` is problematic - I think the cleanest resolution is to only allow for a single call

anandthakker:
@yoshuawuyts @toddself Hey, just a heads up -- I've lost the reference to where I read about this, but I distinctly remember reading that `Object.freeze` has some pretty bad performance characteristics in some browsers (maybe Safari? not sure).  Would you consider guarding this with a check for `process.env.NODE_ENV === 'production'`, so that it could be easily stripped out of a production build?

anandthakker:
>  the fact that a single action can trigger both an effect and reducer is problematic - I think the cleanest resolution is to only allow for a single call

@yoshuawuyts interesting; could you say a bit more about this?   If I'm understanding you correctly, this would effectively remove "actions" as a concept independent of effects and reducers--is that right?

yoshuawuyts:
---

> Object.freeze has some pretty bad performance characteristics in some browsers (maybe Safari? not sure).

Oh yeah, that's a good call. We could def add an argument that can be passed on `start` - I've found it's generally  not a good idea to have env var switches live within packages.

---

>  If I'm understanding you correctly, this would effectively remove "actions" as a concept independent of effects and reducers--is that right?

Ummm, maybe? The way I view actions is as a single trigger on any of the stores - a `1:1` mapping if you will; quite similar to a function call. Having it trigger multiple things at once seems... well confusing haha

---

Hope this answers your questions!

anandthakker:
> The way I view actions is as a single trigger on any of the stores - a 1:1 mapping if you will; quite similar to a function call. Having it trigger multiple things at once seems... well confusing haha

Interesting.  In my work, I've been thinking of "actions" as a separate abstraction, basically representing any "verb" within the system (some of which map directly to user actions like `selectSomeItem` or `cancelSomething`, others of which are more internal like `fetchSomeData` or `updateBlahStatus`).  Coming from that point of view, I feel that it can sometimes make sense for an action to trigger multiple things, because I think there isn't always a 1:1 mapping between "verbs" and stores.  But if there's going to be a 1:1 mapping, then would it be simpler to eliminate actions altogether, in favor of an API like `(state, effects, stores) => { ... stores['namespace:reducerName'](args) ... }`?

toddself:
I have it set up similarly but rather use effects to trigger multiple reducers to react to input. 

The user might perform some action which causes several parts of the state to update, so i trigger an effect which in turn triggers all the single modification reducers to fire that are necessary.

anandthakker:
Yeah, I can see how that makes sense, although again, if there's going to be a 1-1 mapping between actions and effects||reducers, then it seems to me like the "actions" concept is unnecessary...

toddself:
It's more of one-to-many.  One effect can trigger several reducers, but 
a reducer only ever changes one part of the state and will never call another
reducer.

toddself:
Here is an example:
https://github.com/toddself/boxcar/blob/master/effects/move.js

anandthakker:
> It's more of one-to-many.  One effect can trigger several reducers, but a reducer only ever changes one part of the state and will never call another reducer.

effects-reducers is 1-to-many, but I thought @yoshuawuyts was proposing above that **actions**-{reducers or effects} should be 1-1

yoshuawuyts:
> (...) actions-{reducers or effects} should be 1-1

Yup you're both right. One action can trigger either one effect or one reducer; effects can emit multiple actions. Not the current behavior, but think it would make more sense

anandthakker:
> Yup you're both right. One action can trigger either one effect or one reducer; effects can emit multiple actions. Not the current behavior, but think it would make more sense

@yoshuawuyts I can definitely see your point on the clarity/simplicity of doing this, but it still leaves me with the question about what value is then being added by the "action" concept?  Since each action would simply map directly to either a reducer or effect, they really are equivalent to function calls (like you said above) -- so at that point, why is the indirection through `send(...)` necessary or preferable to an API where you just call methods directly?

yoshuawuyts:
> why is the indirection through send(...) necessary or preferable to an API where you just call methods directly

Because multiplexing. A function call requires knowledge of which function you're calling, and importing it explicitly from the location it's declared. By using a message bus, function calling becomes centralized and thus more loosely coupled. This enables better tracing, logging and upgrading parts of the application without touching other parts. It also clearly signals that it's a different method of calling functions; one where callbacks or returning eventual values doesn't fly - thus making for a better enforcement of the unidirectional data flow (despite there always being ways of cheating around it).

anandthakker:
> A function call requires knowledge of which function you're calling, and importing it explicitly from the location it's declared. By using a message bus, function calling becomes centralized and thus more loosely coupled. This enables better tracing, logging and upgrading parts of the application without touching other parts.

Ah, yeah: I totally agree on the value of a message bus internally.  What I was thinking/wondering about was not plain function calls, but rather an API where instead of getting `send` as a parameter, views (and effects, etc.) would get something like `{ effects, stores }`, and where `stores['thing1'](data)` or `effects['thing2'](data)` would be the exact equivalent of `send('thing1', data)` or `send('thing2', data)`...

> It also clearly signals that it's a different method of calling functions; one where callbacks or returning eventual values doesn't fly

... but I hadn't thought of this point, which I can see being a downside of the thing I'm describing above.  I guess now that I've written it out more explicitly, the thing I was suggesting adds more magic and more internal machinery for little gain.

Thanks for engaging with my speculations and questions!

toddself:
@yoshuawuyts agreed on the fact that `Object.keys(reducers).concat(Object.keys(effects))` should be a unique set otherwise it makes it hard to reason about order of operations when you trigger something that fires and effect and a reducer.

yoshuawuyts:
@toddself Given that this is moved to https://github.com/yoshuawuyts/choo/tree/browser-tests and `barracks` I think it's safe to close right? :sparkles:

toddself:
💯

PR #48, opened 2016-06-16
Clarify documentation on id overwrite

toddself:
Was playing around today and couldn't find my element that I was lazily marking with an `id` after it was inserted into the DOM.

Noticed the call to `setAttribute('id', 'choo-root')` and figured this should get called out in the docs probably since it could cause confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,9 @@ const mainView = (params, state, send) => choo.view`
 `
 ```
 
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
ahh yeah; that's annoying - thanks for the docs fix! If you've got a better suggestion on how to handle the ID juggling it'd be more than welcome :sparkles:

Issue #49, opened 2016-06-18
Multiple choo components on a page


toddself:
Just was futzing around with testing and realized that since choo controls it's domain via the addition of `#choo-root` on the element and the fact that everything is initialized and kept in the choo closure, it makes it impossible to have more than one choo component per page.

IMO it would be ideal if you could create multiple choo-owned objects per page, each with their own model and part of the DOM (you could add a data attribute with a randomized id and move choo into a "factory" of sorts (either through `new` or returning an `Object.create` etc)

timwis:
I recall from the code that `#choo-root` is [only the default name of the container](https://github.com/yoshuawuyts/choo/blob/master/index.js#L62). You should be able to have multiple components on a page via:

```javascript
const app1 = choo({ name: 'app1' })
const app2 = choo({ name: 'app2' })
```

toddself:
Ah. Somehow missed that.  It means you also have to namespace your model it seems? Thanks @timwis

timwis:
Nope - I don't believe `name` and `namespace` are connected.

toddself:
I mean namespacing for issues like this:

```js
const choo = require('choo')

const app1 = choo({name: 'app1'})
const app2 = choo({name: 'app2'})

app1.model({
  state: {
    data: 'app1'
  },
  reducers: {
    update: (action, state) => ({data: action.value})
  }
})

app2.model({
  state: {
    data: 'app2'
  },
  reducers: {
    update: (action, state) => ({data: action.value})
  }
})

function app1main (params, state, send) {
  return choo.view`<button onclick=${(evt) => send('update', {value: 'app1main'})}>${state.data}</button>`
}

function app2main (params, state, send) {
  return choo.view`<button onclick=${(evt) => send('update', {value: 'app2main'})}>${state.data}</button>`
}

app1.router((route) => [
  route('/', app1main)
])

app2.router((route) => [
  route('/', app2main)
])

const tree1 = app1.start()
const tree2 = app2.start()

document.body.appendChild(tree1)
document.body.appendChild(tree2)
```

If you click on the second button, it changes the state of the first button.  Just adding a namespace to each new call to `choo#model` and the `send` events doesn't rectify this situation. 

```js
const choo = require('choo')

const app1 = choo({name: 'app1'})
const app2 = choo({name: 'app2'})

app1.model({
  state: {
    data: 'app1'
  },
  reducers: {
    update: (action, state) => ({data: action.value})
  }
})

app2.model({
[...]

PR #52, opened 2016-06-19
Allow for naming the rootId with no options

toddself:
Right now if you want to name your `rootId` something different, but don't pass
in any options, it doesn't change the name of your `rootId` (see confusion in
github #49.)

This would allow someone to use

```js
const app = choo()
app.start('myRootNode')
```

Rather than having to use

```js
const app = choo()
app.start('myRootNode', {name: 'myRootNode'})
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,7 +54,7 @@ function choo () {
   // start the application
   // (str?, obj?) -> DOMNode
   function start (rootId, opts) {
-    if (!opts) {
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
ahh dammit - great catch!

yoshuawuyts:
released as `v2.2.2` :sparkles:

PR #53, opened 2016-06-19
Fix assert call for subscriptions

toddself:
This actually works but you don't get the assertion message, you just get the `false == true` message.


yoshuawuyts:
haha, neato - good catch again ^__^

yoshuawuyts:
released as `v2.2.2` :sparkles:

PR #64, opened 2016-06-21
Update README.md

pguth:
Can't set namespace under `effects`.

Either (pullrequest)
```js
const app = choo()
app.model({
  namespace: 'app',
  subscriptions: [
    (send) => setInterval(() => send('app:print', { payload: 'dog?' }), 1000)
  ],
  effects: {
    print: (action, state) => console.log(action.payload)
  }
})
```
or
```js
const app = choo()
app.model({
  subscriptions: [
    (send) => setInterval(() => send('print', { payload: 'dog?' }), 1000)
  ],
  effects: {
    print: (action, state) => console.log(action.payload)
  }
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,11 +289,12 @@ An example subscription that logs `"dog?"` every second:
 ```js
 const app = choo()
 app.model({
+  namespace: 'app',
   subscriptions: [
     (send) => setInterval(() => send('app:print', { payload: 'dog?' }), 1000)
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
oops, yeah that's correct - bit of a remnant from before namespaces were introduced. Thanks!

pguth:
:satisfied:

PR #72, opened 2016-06-22
WIP: browser testing

toddself:
Wanted to get some feedback on this.

Right now the tests will NOT complete in IE nor Safari due to the fact they're using keyboard events to trigger the tests, however, karma adds some baggage here and I want to make sure we're all ready for it.

The upside is that it will run the tests via SauceLabs in a panoply of browsers for us, so we can prove the stuff works (with some minor compiling down to es2020) in those browsers.

Also, this will require @yoshuawuyts to sign up for a free Sauce Labs account and configure Travis thusly. (Just making sure the username and access_key are in the env variables section of travis where the karma config wants them to be.)

Obviously this needs tests too, but this at least lets us start writing them.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,114 @@
+'use strict'
+
+const fs = require('fs')
+
+module.exports = function (karma) {
+  const glob = 'tests/browser/*.js'
+  const files = [{pattern: glob}]
+
+  const customLaunchers = {
+    SL_Edge: {
+      base: 'SauceLabs',
+      platform: 'Windows 10',
+      browserName: 'microsoftedge'
+    },
+    SL_Firefox: {
+      base: 'SauceLabs',
+      browserName: 'Firefox'
+    },
+    SL_Chrome: {
+      base: 'SauceLabs',
+      browserName: 'Chrome',
+      platform: 'Windows 10'
+    },
+    SL_Safari: {
+      base: 'SauceLabs',
+      browserName: 'safari',
+      platform: 'OSX 10.11'
+    },
+    SL_IE11: {
+      base: 'SauceLabs',
+      browserName: 'internet explorer',
+      version: '11'
+    },
+    SL_IE10: {
+      base: 'SauceLabs',
+      browserName: 'internet explorer',
+      version: '10'
+    },
+    SL_IE9: {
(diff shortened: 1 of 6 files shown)


yoshuawuyts:
Cool; I've added the Sauce credentials to travis. Now all that's needed is to add this snippet to `.travis.yml`:
```yml
addons:
  sauce_connect: true
```
Hope that works; tests are looking good hey!

toddself:
So! I committed, but it seems like the sauce credentials aren't in travis? 

When I start it says "no environment variables" so it's trying to load them from disk.

I ran them locally with my credentials from boxcar though and it's failing on IE 10 due to an issue with `const` declarations.

It looks like some-how sheet-router is NOT having it's `const` declarations converted to `var` via es2020, but only sheet-router.

This is very odd -- I gotta run to my mother-in-law's birthday but I've pushed this out if you want to take a look and see if you can figure it out @yoshuawuyts

toddself:
Here is a [gist](https://gist.github.com/toddself/ce425e366c2fd47017e0c2fc244ad0d6) with the results of the browserify output that the test is using.  You can see the very top of the file where choo is placed, it's all `var` until you get to `hash.js` from sheet-router where it converts back to `const`.

This could be a bug in a bunch of places though :(

karma.conf.js
```diff
@@ -0,0 +1,114 @@
...
+      [glob]: ['browserify']
+    },
+    reporters: ['dots'],
+    port: 9876,
+    colors: true,
+    logLevel: karma.LOG_INFO,
+    files: files,
+    browserify: {
+      debug: true,
+      transform: ['sheetify/transform', 'es2020']
```
yoshuawuyts: ah yeah, `es2020` should be under the `"global"` keyword here, as `transform` only applies to code declared in the repo :sparkles:

yoshuawuyts:
ahh yeah interesting; perhaps the credentials should be made public? But then I probably need to encrypt them 😪 - alright, gonna ping Travis for this and see what they say

karma.conf.js
toddself: D'oh!

Totally missed that one. :(. Thanks!

toddself:
Closing in favor of #86

PR #73, opened 2016-06-23
Fix mailbox example

zigomir:
I was playing with `choo` examples and found a bug while reloading page when specific message was selected. Example: http://localhost:8080/inbox/2

Problem was that `bankai` was returning `bundle.js` as `text/html`. I know my fix isn't good, bit I couldn't find a way to tell `bankai` how to return `js` file from whatever route correctly. It works well for on first level: http://localhost:8080/inbox

Probably I should report this to bankai repo, but I'm not sure. If that's the case tell me and I'll take a better look there.

Also adding missing `tachyons` dependency.

### DIFF
--- a/examples/mailbox/package.json
+++ b/examples/mailbox/package.json
@@ -15,6 +15,7 @@
   "license": "ISC",
   "dependencies": {
     "css-wipe": "^4.2.1",
-    "dateformat": "^1.0.12"
+    "dateformat": "^1.0.12",
+    "tachyons": "^4.0.0-beta.33"
   }
 }
--- a/examples/mailbox/server.js
+++ b/examples/mailbox/server.js
@@ -11,13 +11,14 @@ server.listen(PORT, () => process.stdout.write(`listening on port ${PORT}\n`))
 function createRouter () {
   const router = serverRouter('/404')
 
+  const js = bankai.js(browserify, require.resolve('./client.js'))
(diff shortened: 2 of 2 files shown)


yoshuawuyts:
Actually I wonder if bumping `bankai` would do the trick - merged this a couple of days ago https://github.com/yoshuawuyts/bankai/commit/52825f6bd8d90528fe07a5da3a05c902752b9833

zigomir:
I have `bankai@2.0.5`...

zigomir:
I'm looking at `bankai` now and have troubles running tests. `js returns data` is failing for me.

I know this isn't a scope of `choo`, but I'm thinking it would be easier (at least for me) to just use different server for this particular example. What would you do?

zigomir:
What is most strange (to me), is that I tried to fix example with [wzrd](https://github.com/maxogden/wzrd) with `--pushstate` option and also with [httpster](https://github.com/SimbCo/httpster) (also with `--pushstate` option) and even [pushstate-server](https://github.com/scottcorgan/pushstate-server) and none will work :/

Maybe I'm doing something wrong, but whenever you want to get javascripts for route /:mailbox/:message route you don't get it :/ This can't be client side problem, right? Pretty confused right now.

yoshuawuyts:
Ehhh, yeah alright - merging this as it solves a problem, despite it not being the cleanest. Will def have to dig into this for `bankai` as it should be fixed (I'll be working on that after v3). Thanks again! :D

PR #76, opened 2016-06-23
router: fix arguments bug

### LINKED ISSUE #75: bug: passing default path in router causes error
As per https://github.com/yoshuawuyts/sheet-router/issues/21 when passing a default path to the router, an error is thrown:
```txt
assert.js:59Uncaught AssertionError: createTree must be a function
```
This only happens in `choo`. Should probably be an argument order error

yoshuawuyts:
Closes #75. Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -179,9 +179,9 @@ function choo () {
   }
 
   // register all routes on the router
-  // [obj|fn] -> null
-  function router (cb) {
-    _router = sheetRouter(cb)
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Not sure if it's better to do it this way - as it creates an array, or just to do an `if` statement and select arguments based on that.

yoshuawuyts:
ps if people think this works, please go ahead and release - reckon I might be offline for the next 16hrs or so

PR #90, opened 2016-06-27
Call `http.del` in httpDelete function

zigomir:


yoshuawuyts:
hah, good catch! Thanks!

PR #91, opened 2016-06-27
`document` not used in form snippet

zigomir:
GH diff looks weird, I only removed `const document = require('global/document')` :)


yoshuawuyts:
@zigomir yeah it def does - we need to look into that. In the mean time: change looks good - thanks again! :sparkles:

Issue #42, opened 2016-06-07
3.0 release log


yoshuawuyts:
We're probably going to run some breaking changes, for the sake of a better API. These changes should get us around `5kb` optimized.

- [x] move `choo.view` out to `require('choo/html')` #71 | pr #103
- [x] streamline view API #35 | pr #111 
- [x] higher order functions #34 | pr #104
- [x] create lifecycle hooks #1 | feature addition in dependency covered by semver
- [x] implement state hooks #15 | pr #104 
- [x] add yo-yoify #3 | pr #110
- [x] rename "app" namespace #82 | pr #111 
- [x] enable browser testing | pr #86
- [x] propagating actions creates infinite loop #114 | pr #104 

When everything is merged:
- [x] validate all tests pass
- [x] validate all examples work
- [x] validate all docs are updated
- [x] write changelog

Please do comment if I've missed anything; I'm not a fan of breaking changes, but if it needs breaking this is the time to speak up. Cheers!

toddself:
We should probably add 

[ ] enable browser testing

To this list :)

yoshuawuyts:
@toddself feel free to update the release list as you like; I just happened to create it - anyone with commit rights should be able to do that ^__^

PR #95, opened 2016-06-29
Extract html into separate module

### LINKED ISSUE #3: browserify transform
add `yo-yoify` transform so values `hyperx` can be stripped which saves us another `3kb`; down from `7.2kb` to `4.6kb`

### LINKED ISSUE #60: Unbundle the http module?
Wanted to get your thoughts on whether the `http` module still makes sense to bundle with choo? As opposed to pointing folks to it in the readme as a "recommended" module. But if you don't need to make http requests, it isn't really relevant for you.

toddself:
This should satisfy splitting out for yo-yoify support.

Another 3.0 roadmap fix (see #42)

Fixes #3
Fixes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Let's create an input box that changes the content of a textbox in real time.
 First we import `choo` and create a new instance:
 ```js
 const choo = require('choo')
+const html = require('choo/html')
 const app = choo()
 ```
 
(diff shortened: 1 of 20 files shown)


yoshuawuyts:
@toddself could you create a separate `3.0` release branch and PR against that? Means I can PR some of my stuff agains that too and we get to test stuff out before releasing. Means we can put out the whole load of the 3.0 release in a single go. Thanks! :sparkles:

toddself:
@yoshuawuyts gladly

PR #102, opened 2016-06-30
Added bower.json file as a dependency

sibStephen:
Added Bower.json file to install bower specific packages.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,18 @@
+{
+  "name": "choo",
+  "description": "A 7kb framework for creating sturdy frontend applications",
+  "main": "index.js",
+  "authors": [
+    "N"
+  ],
+  "license": "MIT",
+  "keywords": [
+    "client",
+    "frontend",
+    "framework",
+    "minimal",
+    "composable",
+    "tiny"
+  ],
+  "homepage": "https://github.com/sibStephen/choo"
+}


yoshuawuyts:
Hey @sibStephen, thanks for your PR! :tada:

I've got to admit I dislike `bower` with a burning passion so I'm not going to merge this; I want to __strongly__ discourage people from picking it over `npm`. If `choo` needs to be backported to a legacy project that happens to use `bower`, I'd recommend creating a new project that isn't littered with cruft and start fresh - we support hooking into arbitrary DOM nodes right now so parts could be ported; and soon we'll probably document how to create multiple apps that can be combined into a single larger application.

Thanks for understanding; I hope you're enjoying `choo` as much as we're enjoying building it! :grin:

jonkoops:
@sibStephen There really shouldn't be a reason anymore why you wouldn't be able to use npm for this. If you really want to use Bower I would recommend you use the URL to the GIT repository like so:

`$ bower install --save https://github.com/yoshuawuyts/choo.git`

sibStephen:
Yes @jonkoops  and @yoshuawuyts  i totally understand, `npm`  is efficient when compared to `bower` but yeah i thought that it would serve some of the users who might not be well versed with the npm usage, anyways thanks for the advice @yoshuawuyts  and yes i'm using choo :)

PR #103, opened 2016-06-30
Extract html into separate module

### LINKED ISSUE #3: browserify transform

### LINKED ISSUE #60: Unbundle the http module?

toddself:
This should satisfy splitting out for yo-yoify support.

Another 3.0 roadmap fix (see #42)

Fixes #3
Fixes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Let's create an input box that changes the content of a textbox in real time.
 First we import `choo` and create a new instance:
 ```js
 const choo = require('choo')
+const html = require('choo/html')
 const app = choo()
 ```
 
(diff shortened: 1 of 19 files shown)


yoshuawuyts:
This is the best :tada::tada::tada::tada:

Issue #105, opened 2016-06-30
expose umd build on npmcdn


yoshuawuyts:
[From Twitter](https://twitter.com/nolanlawson/status/748648096565858304) - it'd be a good idea to add `npmcdn` support. All we need to do is create an (optimized) standalone build whenever `npm publish` is run that's written to the `umd/` directory ([ref](https://npmcdn.com/#/)).

This is useful for people using jsfiddle and the like. Think it's fair enough to add this as it's little to no overhead. Thanks!

yoshuawuyts:
Example https://github.com/joshwnj/react-visibility-sensor/blob/master/package.json#L9

PR #108, opened 2016-07-01
prepare umd build before publish

just-boris:
* added browserify build on prepublish hook
* added example how to use UMD build
* updated package.json to exclude unnecessary files from npm package (examples and test)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,10 @@
     "test:browser:local": "standard && npm run deps && NODE_ENV=test zuul --local 8080 -- tests/browser/*",
     "test:cov": "npm run test:server:cov",
     "preversion": "if [ ! -z $SKIP_TEST ]; then npm run test:browser; fi",
-    "test": "npm run test:server"
+    "test": "npm run test:server",
+    "build:dev": "NODE_ENV=development browserify index.js --standalone=choo -t envify > dist/choo.js",
+    "build:min": "NODE_ENV=production browserify index.js --standalone=choo -t envify -g unassertify -g uglifyify | uglifyjs > dist/choo.min.js",
+    "prepublish": "mkdir -p 'dist/' && npm run build:dev && npm run build:min"
   },
   "repository": "yoshuawuyts/choo",
   "keywords": [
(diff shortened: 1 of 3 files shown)


just-boris:
fixes #105

package.json
```diff
@@ -10,7 +10,11 @@
     "test:browser": "standard && npm run deps && NODE_ENV=test zuul tests/browser/*",
     "test:browser:local": "standard && npm run deps && NODE_ENV=test zuul --local 8080 -- tests/browser/*",
     "test:cov": "npm run test:server:cov && npm run test:browser",
-    "test": "npm run test:server && npm run test:browser"
+    "test": "npm run test:server && npm run test:browser",
+    "start": "http-server",
```
yoshuawuyts: I reckon this script should best be scoped to `examples/vanilla`

yoshuawuyts:
Wooh, thanks! - Just left one comment; rest looks brilliant to me! :sparkles:

package.json
just-boris: It uses files from `dist` directory. So we need to serve from project root.

yoshuawuyts: A relative path can be passed as the first argument to `http-server` to determine the location served:
```sh
$ http-server ../../
```

just-boris:
I realised that is it possible to open the example, as any local html-file with `file://` protocol.
No need to keep http-server at all

yoshuawuyts:
Could you remove the `.gitkeep` file?

just-boris:
@yoshuawuyts now we are writing output file using standard `>` operator.
`.gitkeep` is used to create this directory after `git clone`, because `>` fails if target directory doesn't exist.

Do you know another solutions how to automatically create the `dist` folder if it doesn't exist?

yoshuawuyts:
Gehe, unix tools to the rescue!
```sh
$ mkdir -p 'dist/'
```

just-boris:
Done.

Added this as first step of `prepublish` script.

timwis:
This is awesome! One thing as a heads up: 3.0 separates `choo.view` into `require('choo/html')`, a separate file from `index.js`, so I imagine this script may need to be updated to publish both when that's ready.

yoshuawuyts:
@timwis ooh; good call hey

yoshuawuyts:
I've merged the commits into #159 and rebased onto master to resolve the merge conflicts. Heaps thanks!

just-boris:
Good. Anyway I was glad to help you :+1:

PR #109, opened 2016-07-01
only run the tests when trying to make a new version

toddself:
So!

I always forget that `prepublish` runs on `install` when you run install from a cloned repo. Which means anyone doing an `npm install` in choo to do dev works would have to run the tests. 

I disagree with this concept, but that [is a very well traveled road](https://github.com/npm/npm/issues/3059) so instead of running on publish, it runs when you try to generate a new version with `npm version`.

Not the best solution, but definitely better than forcing devs to sit through a test in order to work on choo.

You  CAN skip the test with `SKIP_TEST=true npm version [patch|minor|major]`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,9 @@
     "test:server:cov": "standard && npm run deps && NODE_ENV=test istanbul cover tests/server/*",
     "test:browser": "standard && npm run deps && NODE_ENV=test zuul tests/browser/*",
     "test:browser:local": "standard && npm run deps && NODE_ENV=test zuul --local 8080 -- tests/browser/*",
-    "test:cov": "npm run test:server:cov && npm run test:browser",
-    "test": "npm run test:server && npm run test:browser"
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Looking great! - Could you perhaps add a little `## Contributing` section to the readme so people can be made aware of this? Just for now so that information isn't lost. Heaps thank! :sparkles:

toddself:
Done!

package.json
```diff
@@ -10,7 +10,8 @@
     "test:browser": "standard && npm run deps && NODE_ENV=test zuul tests/browser/*",
     "test:browser:local": "standard && npm run deps && NODE_ENV=test zuul --local 8080 -- tests/browser/*",
     "test:cov": "npm run test:server:cov && npm run test:browser",
```
yoshuawuyts: haha, was wondering why tests were still failing - might need to update this line too, as this is what runs through Travis 😋

PR #104, opened 2016-06-30
architecture: use barracks

yoshuawuyts:
Changes
=======
- add hook handlers
- add effect composition
- add noFreeze option

Commits
=======
- fixup! fix toString()
- fixup! use assert instead of throw
- fixup! fix reducers
- fixup! work on done callbacks
- fixup! more callback work
- fixup! woooh fixed effects errs
- fixup! wrap up stuff

Todo
====
- documentation
- update existing tests

Related to #42 #34 #15 Thanks!

Issue #82, opened 2016-06-25
Rename "app" model namespace?


timwis:
Per convo in IRC, the [`app` model](https://github.com/yoshuawuyts/choo/blob/master/index.js#L202-L230) really only deals with `location` (the route), potentially the [`params`](https://github.com/yoshuawuyts/choo/issues/35), and subscriptions to hash, history, anchor links - all to do with navigating. @yoshuawuyts and I were discussing the idea of renaming the namespace to something like `location`, `navigator`, or `url`. Yosh pointed out `location` has a parallel to `window.location` so may be the best choice.

Thoughts?

yoshuawuyts:
Yup, I like this. I'd keep `state.params` as separate though, as it bears similarities to `react-router` (`this.props.params`) and is easier to access which makes for better UX

yoshuawuyts:
Closing as 3.0 is imminent; wanna make sure all issues are taken care of. Releasing soooon™ :sparkles:

Issue #35, opened 2016-06-02
update view API to be (state, prev, send)


yoshuawuyts:
Since `v2.1.4` (oops, semver) we're now also passing the `oldState` down to views. The `view` api now looks like:
```js
const view = (params, state, send, oldState) => choo.view`
  <main>${params.foo} and ${state.bar}</main>
`
```
Instead I reckon it might be nicer to have:
```js
const view = (state, oldState, send) => choo.view`
  <main>${state.params.foo} and ${state.bar}</main>
`
```
This would definitely be a breaking change, but actually not that hard to fix (e.g. we can provide a regex heyyyy). What do people think?

timwis:
:+1: to reordering the params. Would it make sense to put `params` inside the `app` namespace like `location`?

ahdinosaur:
> Since `v2.1.4` (oops, semver) we're now also passing the `oldState` down to views.

[context](https://github.com/yoshuawuyts/choo/pull/27#issuecomment-222549654) for anyone like me who wasn't aware.

@yoshuawuyts to be honest i'm not quite sold on `oldState` being necessary in views, but happy to follow along with how `choo` develops. :smile:

ahdinosaur:
or i'm being silly, a similar project [`vdux` also passes in previous state](https://github.com/vdux/vdux#components), as do many other projects.

contra:
```js
const view = ({ params, state, oldState}, send) => choo.view`
  <main>${params.foo} and ${state.bar}</main>
`
```

🙏

kristoferjoseph:
```js
const view = (state, send) => choo.view`
  <main>${state.params.foo} and ${state.bar}</main>
`
```

yoshuawuyts:
@kristoferjoseph but where would the `oldState` go? Right now I'm leaning towards:
```js
const html = require('choo/html')
const view = (state, prev, send) => html`
  <main>${state.params.foo} and ${state.bar}</main>
`
```
We could maybe get super smart and do a `Function.prototype.length` check so that it's passed oldState, but only if it cares about it:
```js
const mainView = (state, send) => html`
  <main>${state.params.foo} and ${state.bar}</main>
`
const diffView = (state, prev, send) => html`
  <div>${prev.foo} become ${state.foo}</div>
`
```

This would have the downside that omitting `send()` would make it questionable; but I reckon that's a bit of an edge case:
```js
// prev.foo is undefined
const diffView = (state, prev) => html`
[...]

PR #111, opened 2016-07-01
Change view arguments

yoshuawuyts:
Relevant commit are 	d0788fc and a3bb4f3. Builds on top of #104 

- creates a named `send()` function for views; e.g. caller in `onAction` becomes the name of the view
- changes order of arguments as per #35 
- changes internal `app` namespace to `location` as per #82 

Needs testing and rebase obv, but this should provide a good idea of stuff

### DIFF
--- a/examples/http/views/main.js
+++ b/examples/http/views/main.js
@@ -1,6 +1,6 @@
 const html = require('../../../html')
 
-module.exports = function (params, state, send) {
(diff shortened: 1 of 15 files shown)


yoshuawuyts:
Tests pass, examples work - merging into `v3` :sparkles:

PR #118, opened 2016-07-03
Use relative links to examples

sotojuan:
Cleans up the links to example directories by making them relative.

PS: Someone should look into hosting the mailbox example online so we can give that a real link! I might do it myself later.


yoshuawuyts:
LGTM! - Thanks for this :sparkles: - and you're right, having examples online would be heaps nice

PR #120, opened 2016-07-05
deps: bump to barracks 8

yoshuawuyts:
Updates to barracks 8, which isn't published yet - tests pass locally. Once v8 is release we should bump the version so it can be merged. :sparkles:


yoshuawuyts:
Welp, tests pass - mergin!

Issue #81, opened 2016-06-24
Batched updates


anandthakker:
Depending on the store/action design of an app, batching updates can be a pretty useful/important perf optimization.  @yoshuawuyts what are you thoughts on the best `choo`-like way to handle this?  Would you be interested in adding a hook that would make this possible in some app-wide way (like maybe a way to provide 'middleware' function through which updated states were passed before being emitted to the view-rendering)?  Or would you prefer for people to handle this by simply throttling individual view functions?

yoshuawuyts:
Yeah I've been considering this; combined with documenting best practices on thunking views and caching routes (https://github.com/yoshuawuyts/sheet-router/issues/17) it would def be worthwhile to put an upper cap on flushing updates.

In the past for `virtual-dom` applications I used [yoshuawuyts/virtual-raf](https://github.com/yoshuawuyts/virtual-raf) to flush updates at 60fps; it could def be worthwhile exploring this for `choo` too - especially the interaction with webworker rendering would become interesting (https://github.com/maxogden/yo-yo/issues/22).

I think this should happen after the `3.0` release though; as it's a minor tweak without practical API changes. Does this answer what you meant? Cheers!

yoshuawuyts:
Created https://github.com/yoshuawuyts/nanoraf yest; might add this. Talked about this on IRC, and considered not doing batched updates; but realized yesterday that sometimes something like a higher order effect might trigger a whole slew of updates in rapid succession because stores are separated - there'd be no way of throttling that so we should prob def do this. If not today, then soon :sparkles:

yoshuawuyts:
Created a new branch: https://github.com/yoshuawuyts/choo/tree/raf which uses `nanoraf` - it's based off the `3.0` branch and adds full batched updates.

This won't be part of `3.0` however as we probably need to test it a little, and I'd rather put out `3.0` before then - but it's non-breaking so can de make it `3.1` or smth :sparkles: Only difference that occurs is that older (`<IE10`) browsers will need a shim. I don't think that's bad enough to call a breaking change as it will realistically affect less than 1% of browsers out there, probably less for this lil choo thing we got here.

Anway, it can be played with now; keen to hear your thoughts on it.

anandthakker:
@yoshuawuyts I haven't given it an actual test drive, but from a read of the code the approach looks great to me!  I have a couple of thoughts but I'll post them as issues over in nanoraf.

yoshuawuyts:
[...]

PR #135, opened 2016-07-06
index: add raf guard

### LINKED ISSUE #99: improve perf
Hi,

I'm testing `choo` against others and really wish it perform great but it turn out to be really slow there.
so I think I should ask you first to ensure that I didn't miss something which mess the results.

todomvc-pref-choo

here's the code https://github.com/rabbots/todomvc-perf

Thanks

yoshuawuyts:
Adds a `requestAnimationFrame` loop, so unbound state mutations don't thrash the UI thread. The more boring cousin to isolated mutable updates (which I might gist up soonish, perhaps). Means it'll only run on IE10+ without a shim attached on the `window` (e.g. < 1% browsers are affected).

Provides a 10000x perf improvement in some scenarios, or something like that.

Closes #99 #81. Thanks! :sparkles:

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ const href = require('sheet-router/href')
 const hash = require('sheet-router/hash')
 const hashMatch = require('hash-match')
 const barracks = require('barracks')
+const nanoraf = require('nanoraf')
 const assert = require('assert')
 const xtend = require('xtend')
 const yo = require('yo-yo')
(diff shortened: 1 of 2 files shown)


toddself:
Are you a 10000x programmer now? If I use choo in your project does that make me one as well?!

yoshuawuyts:
@toddself  you're an ∞ programmer

yoshuawuyts:
Should probs release this together with #108 as a minor version

PR #139, opened 2016-07-07
updated readme example for subscriptions and send method

traducer:
as described in https://github.com/yoshuawuyts/choo/issues/137


yoshuawuyts:
Oh nice, thanks for adding that. Real sorry about the callback not being immediately clear - def need to brush up that part of the docs a bit. :sparkles:

PR #143, opened 2016-07-08
Correct model namespace example

timwis:
The current docs use `myTodos` in the code but `todos` in the narrative. This PR makes the narrative match the code and uses `todos` and `items` to differentiate between the namespace and the array in the state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,17 +236,17 @@ namespaced or not. Namespacing means that only state within the model can be
 accessed. Models can still trigger actions on other models, though it's
 recommended to keep that to a minimum.
 
-So say we have a `myTodos` namespace, an `add` reducer and a `todos` model.
+So say we have a `todos` namespace, an `add` reducer and a `todos` model.
 Outside the model they're called by `send('todos:add')` and
-`state.todos.todos`. Inside the namespaced model they're called by
-`send('todos:add')` and `state.todos`. An example namespaced model:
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Oh yeah, lgtm!

README.md
```diff
@@ -236,17 +236,17 @@ namespaced or not. Namespacing means that only state within the model can be
...
 ```js
 const app = choo()
 app.model({
-  namespace: 'myTodos',
-  state: { todos: [] },
+  namespace: 'todos',
+  state: { items: [] },
   reducers: {
-    add: (data, state) => ({ todos: state.todos.concat(data.payload) })
+    add: (data, state) => ({ todos: state.items.concat(data.payload) })
```
slaskis: Is the key of the returned object the namespace? 

I'd expect it to be the new state, so instead of `({ todos: state.items.concat(data.payload) })`

I'd thought it would be `({ items: state.items.concat(data.payload) })`.

timwis: Yikes! Good catch

yoshuawuyts: I'd rather we not force push to master; might even disable that from settings - feel free to just push a new commit tho, it's just docs and the fix is straight forward so no need to run CI :sparkles:

timwis: Forgot that would require a force push, thx!

PR #144, opened 2016-07-08
onAction: prefix views with view

yoshuawuyts:
Prefix name with `view` so `onAction` knows what's up. Thanks! cc/ @timwis


yoshuawuyts:
Tests pass and see no harm in merging - means new version of choo-logger can be pushed ✨

yoshuawuyts:
Published as `v3.0.2`

PR #154, opened 2016-07-10
Stopwatch

traducer:
I feel like a little stopwatch app is a great example for handling state in choo. It also makes use of window.requestAnimationFrame, which will help if anyone plans on using `raf` in their views.

### DIFF
--- a/examples/stopwatch/client.js
+++ b/examples/stopwatch/client.js
@@ -1,0 +1,15 @@
+const choo = require('../../')
+
+const stopwatch = require('./models/stopwatch')
+const mainView = require('./views/main')
+
+const app = choo()
+
+app.model(stopwatch)
+
+app.router((route) => [
+  route('/', mainView)
+])
+
+const tree = app.start()
+document.body.appendChild(tree)
--- a/examples/stopwatch/models/stopwatch.js
+++ b/examples/stopwatch/models/stopwatch.js
@@ -1,4 +1,3 @@
-const xtend = require('xtend')
 const raf = require('raf')
 
 module.exports = {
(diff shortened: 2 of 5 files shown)


examples/stopwatch/client.js
```diff
@@ -0,0 +1,93 @@
...
+  const elapsed = state.stopwatch.elapsed
+  const now = Date.now() - elapsed
+  const start = state.stopwatch.start
+
+  function step () {
+    const elapsed = Math.floor((Date.now() - now))
+    send('stopwatch:update', elapsed)
+  }
+
+  if (start) {
```
yoshuawuyts: Hmm, I think that this piece would be better off in a `subscription`; not as part of a `view` - it _is_ a read-only source of data after all; similar to the `elm` time example http://elm-lang.org/examples/time

traducer:
@yoshuawuyts, I moved the `requestAnimationFrame` logic into a `subscription` which then sends actions to `effects`, hopefully this is what you had in mind. If not let me know. The view should now only reads properties in state and reacts to them.

yoshuawuyts:
haha, forgive me being slightly perfectionist, but there's two comments I'd make:
- could you split this into multiple files? I feel having helper functions, setup, views and models in a single file is a bit much, conceptually
- I'd generally only ever have a click trigger a single `action` - if multiple actions must be called, they'd be best combined inside of an `effect` - maybe this is overdoing it, but if it's in the repo I reckon it's best to reflect how higher order effects were meant to be used.

Looking great tho! :sparkles:

traducer:
haha np, I'll clean it up

traducer:
@yoshuawuyts, I separated the code into different files. Also, my click handlers no longer call multiple actions, only a single action depending on the state. Everything seems to be working as intended.

examples/stopwatch/models/stopwatch.js
```diff
@@ -0,0 +1,34 @@
...
+  state: {
+    elapsed: 0,
+    startTime: 0,
+    start: false,
+    laps: []
+  },
+  reducers: {
+    start: (data, state) => xtend(state, { start: true, startTime: Date.now() - state.elapsed }),
+    stop: (data, state) => xtend(state, { start: false }),
+    update: (data, state) => xtend(state, { elapsed: data }),
```
yoshuawuyts: xtend is not needed here; returning the new state suffices :D

yoshuawuyts: *returning the object diff suffices; the result is extended by choo itself :D

traducer: Wow I did not know that, that's pretty awesome!

I'll update the changes.

traducer:
@yoshuawuyts I updated the example to not use xtend, if there's anything else you feels needs to be changed let me know!

yoshuawuyts:
Nope, this is perfect. Thanks! :D

PR #159, opened 2016-07-12
Scriptsies

### LINKED ISSUE #105: expose umd build on npmcdn

yoshuawuyts:
Builds on top of #108, closes #105. Moved all scripts to the `/scripts` dir so `package.json` stays maintainable. We now have three scripts:
- __test:__ run tests
- __instrument:__ run instrumentation; mostly file size now, perhaps perf and other stuff in the future
- __build:__ create a standalone `umd` bundle

All in valid POSIX `sh`, so if we ever switch away from Travis we should have no issues. Thanks!


yoshuawuyts:
Tests pass, mergin time!

yoshuawuyts:
Or wait... code cov?

yoshuawuyts:
Ahhh, I was a fool - test was exploding - this should do iiiiit

yoshuawuyts:
There we go, `codecov` support is now in :D - bet if there's any issues we can fix 'em later; this PR just targets tooling so I think it should be cool to merge and keep tabs. :D :D

PR #161, opened 2016-07-12
script/build: prune the prod builds

yoshuawuyts:
Makes it so the build script applies all optimizations, rather than just some :sparkles:


yoshuawuyts:
Hurray for `zopfli(1)` :tada:

PR #182, opened 2016-07-20
location: change url on location:setLocation

yoshuawuyts:
I think this should address the issues from https://github.com/yoshuawuyts/choo/issues/85. Hash routes probably won't work well, but in order to resolve that we'd probs have to patch the router (https://github.com/yoshuawuyts/sheet-router/issues/31).

I've tested this on the `mailbox` example and it seems to work well. Passing in two `send()` calls does mean there's a slight delay in the UI, but I think we can get around that by removing `setTimeout()` in `barracks` - which would _probably_ not unleash Zalgo. Maybe it would unleash Zalgo. I think we should leave that up to the consumers to take care of hah.

Anyway, I hope this resolves the outstanding routing issues. Feedback def welcome, as it'd be cool to get this fix right(ish). Cheers!

### DIFF
--- a/examples/mailbox/views/mailbox.js
+++ b/examples/mailbox/views/mailbox.js
@@ -7,6 +7,7 @@ const nav = require('../elements/nav')
 module.exports = function (state, prev, send) {
   return html`
     <main class="mw5 mw7-ns center cf">
+      <button onclick=${goHome}>Home</button>
       ${pathname(state, prev, send)}
       ${nav(state, prev, send)}
       <section class="fl mt4 w-80 db">
@@ -14,4 +15,8 @@ module.exports = function (state, prev, send) {
       </section>
     </main>
   `
+
+  function goHome () {
+    send('location:setLocation', { location: '/'})
+  }
 }
--- a/index.js
+++ b/index.js
@@ -142,8 +142,19 @@ function appInit (opts) {
   const loc = document.location
   const state = { pathname: (opts.hash) ? hashMatch(loc.hash) : loc.href }
   const reducers = {
-    setLocation: function setLocation (data, state) {
-      return { pathname: data.location.replace(/#.*/, '') }
+    updateHref: function (data, state) {
+      return { pathname: data }
+    }
+  }
+  const effects = {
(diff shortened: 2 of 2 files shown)


timwis:
But hashes! I feel like people think of hash routing as something just for old browsers, but what about sites on github pages and other static servers that don't support "overriding the resource and passing it to index.html"? I would advocate for hash routing being treated as a first class citizen - unless there's some other way to use it widespread?

Couldn't we just do someting like `if (opts.hash) window.location.hash = xxx; else window.pushState()..; send(...)`?

yoshuawuyts:
> But hashes! I feel like people think of hash routing as something just for old browsers

No no, not at all. I know it's like the only proper way to do routing in `electron` for example. But we can't get everything in one go, and this at least fixes the issues for people using path. Hashes are not forgotten, but need to wait I think

yoshuawuyts:
You're totally right - this patch _would_ break hash routing - gotta fix it! :D

paul-veevers:
Hey guys, been lurking for a while and loving Choo. Currently building a proof-of-concept third-party embeddable widget (i.e. sits on other people's websites). 

One requirement is that it doesn't change the URL of the website (hash or otherwise). But obviously I still need routing. This has worked perfectly so far:

`
const tree = app.start({
    history: false,
    href: false,
    hash: false,
 });
`
`
send('location:setLocation', {location: '/'}, done);
`

This commit would break this functionality? I.e. it would change the URL.

I can see I will be able to call the `updateHref` effect directly instead? But thought I'd point out a use case that will break 👍 

Look forward to contributing a bit more when I have some breathing room :)

yoshuawuyts:
@paul-veevers I hear you - yeah we should take care of that for sure

yoshuawuyts:
Update: not forgotten about this, life is just a teeny tiny mess at the moment. More than happy to accept PRs against this PR to solve this - might be a few more days still, otherwise. Thanks for understanding :grin:

yoshuawuyts:
Been tinkering with this, and there's almost no way to escape breaking changes - going to close this issue and put out a full router revamp patch

PR #186, opened 2016-07-20
DOC: Add effects example

### LINKED ISSUE #184: Need examples to show how effects work in the README
The README should have a code block showing how an effect might work that calls multiple reducers and to show the arguments to the function provided

toddself:
Add an example to the effects section to show how an effect might have to do
multiple async tasks and to call `done` only when complete.

Fixes #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,27 @@ Examples of effects include: performing
 (server requests), calling multiple `reducers`, persisting state to
 [localstorage][localstorage].
 
+```js
+const http = require('choo/http')
+const choo = require('choo')
+const app = choo()
+app.model({
+  namespace: 'todos',
+  state: { items: [] },
+  effects: {
+    addAndSave: (data, state, send, done) => {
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
LGTM! 👌

timwis:
Hey, just noticed this mutates the `data` object passed to the effect by adding an `id` property. I was under the impression this is bad practise, but I've never write understood why. Figured it's worth bringing up since it's now in the readme - Is it ok or nono?

yoshuawuyts:
@timwis yeah you're right, https://github.com/yoshuawuyts/barracks/blob/master/index.js#L187 is passing data as-is. Even if `Object.freeze()` was used it'd be a problem, as it would straight up `throw`

timwis:
Okay, so just to confirm, we should change this in the readme then, yeah? (A little confused as to which part I was right about lol)

yoshuawuyts:
yeah the example is buggy and we should change it

PR #190, opened 2016-07-21
Update Input Example to latest version of choo.

MattMcFarland:
When I first started using choo, I used the requirebin link first, but then after I wanted to use it in a project it wouldn't work.  After a bit of headache it turned out the requirebin was using version 1.0 but choo 3 was installed.

The very next day, a colleague of mine went through the same problem! 

So I wanted to update this to prevent the pain from happening to anyone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@
 - __very cute:__ choo choo!
 
 ## Demos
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Ahh, yeah nice catch! - thanks so much :grin: :sparkles:

PR #196, opened 2016-07-25
Update link to example

MattMcFarland:
The requirebin demo was also in the example area which I didn't see beforehand.


yoshuawuyts:
Ah nice catch. Thanks! :grin:

PR #198, opened 2016-07-26
Improve README data flow diagram

kemitchell:
This pull request replaces the current README data-flow diagram with a new one. My goals for the new one were to:
1. Bring together more of the vocabulary otherwise spread out in the description of each component. For example, I've tracked verb usage, like "subscriptions emit events" and "actions call reducers". Not just nouns.
2. Depict clearly that "models" are aggregates of subscriptions, reducers, effects, and initial states. They have no role in and of themselves.
3. Retain the cyclical overall look of the current diagram, with stronger hints about direction of flow.
4. Avoid mixing box-drawing and potentially variable-width Unicode symbols, so it renders reliably without alignment issues.
5. Users are people!

This has been very useful in sorting out my only understanding of what choo seems to be going for. That being said, I fully expect I might have tripped up somewhere.

Apart from the mistakes I don't know about, there are a few ways I think this might be improved still:
1. Call out more prominently that asynchrony lives in Effects. At least for me personally, it took a little while to deduce that the server side sneaks in somewhere in the action-effect-action cycle.
2. Clarify that Effects and Reducers share a single namespace unified by `send()`.
3. Work the Model namespace thing in a more teachable way.  I am not entirely convinced that I understand them yet.

Hope this is helpful.

Best,

K

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,14 +206,35 @@ efficiently rendered on the screen.
 In turn when the `views` are rendered, the `user` can interact with elements by
 clicking on them, triggering `actions` which then flow back into the
 application logic. This is the _unidirectional_ architecture of `choo`.
+
 ```txt
- ┌─────────────────┐
- │  Subscriptions ─┤     User ───┐
- └─ Effects  ◀─────┤             ▼
- ┌─ Reducers ◀─────┴──Actions── DOM ◀┐
- │                                   │
- └▶ Router ─────State ───▶ Views ────┘
++----------+     changes trigger       +-------+
+|          |<-------------------------<|       |
+|  Router  |                           | State |     initialize
+|          |                     +---->|       |<-----------------+
++----------+              modify |     +-------+                  |
+  v                              |                                |
+  | renders  +-------------------|--------------------------------|----+
+  v          | Model             |                                |    |
++---------+  |                   ^      call    call              ^    |
+|         |  | Subscriptions  Reducers <----+  +----> Effects  Initial |
+|  Views  |  |       v                      |  |        v      States  |
+|         |  |       |                      |  |        |              |
++---------+  +-------|----------------------|--|--------|--------------+
+  v                  |                      ^  ^        |
+  | update           |     emit           +---------+   | asynchronously
+  v                  +------------------->|         |   | emit
++---------+                               | Actions |<--+
+|   DOM   |>----------------------------->|         |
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
This chart is great! Agree with all the points your pointing out; this def seems to help in establishing a stronger mental model of how all pieces in `choo` relate to each other.

Might turn this into an SVG as it seems ASCII is hitting its limits here hah
screen shot 2016-07-26 at 11 35 05

MattMcFarland:
Have you tried http://asciiflow.com/ ?  

I think it would be perfect for this :)

MattMcFarland:
I've been looking for an excuse to use asciiflow :)

have a look:
http://asciiflow.com/#0B6WuOwtQwe24ZEVhb2t0dy1iZVk

yoshuawuyts:
Perhaps like this? https://gist.github.com/yoshuawuyts/62305bfb8804a9e124818cb3061609a4

YerkoPalma:
Maybe should say that reducers make synchonous changes in the state, as it says that effects are asynchronous?

kemitchell:
@yoshuawuyts Where it rendering mis-aligned?

kemitchell:
Does the first commit, which was pure ASCII, have the same issue?

It might have something to do with the `txt` Markdown fenced block info string. I'm not sure we need one.

yoshuawuyts:
Ohey, yeah the first one looks a lot better - though I'd argue that SVG might be a step up from there even - not too attached to the text-graphs personally hah

kemitchell:
I've amended to use plain ASCII characters again.

I don't know how to convert to SVG efficiently, but I'm certainly not against an SVG diagram.

yoshuawuyts:
Hmm, maybe we should hold this off for a bit until https://github.com/yoshuawuyts/choo/issues/201 is fixed; asked feedback about the new ascii chart, and people didn't seem enthusiastic. Perhaps there's a deeper lying problem in the docs that changing a chart cannot address. Does that sound reasonable?

andineck:
have you tried https://www.draw.io? it has a chrome app that let's you draw offline, it's free. it loads/saves svg directly (without import/export).

yoshuawuyts:
yeah don't think we're gonna merge this. Thanks for the PR regardless tho! - closing :sparkles:

Issue #160, opened 2016-07-12
UMD build should build html.js


timwis:
Per #108, the [new UMD build process](https://github.com/yoshuawuyts/choo/pull/159/files) should also build `choo/html.js` so that you can construct DOM.

marshallswain:
Related to https://github.com/maxogden/yo-yo/issues/38

MattMcFarland:
I added a fix for this but now I'm wondering if we should remove html from choo.js as well because otherwise html is loaded twice, that being said we would end up forcing people to must use html.js though which I dont know if that is what we want.  

Personally I'd prefer separating them.

YerkoPalma:
@MattMcFarland  What do you mean by

>html is loaded twice

MattMcFarland:
So the build script I thought was exposing both html and choo already., otherwise the example wouldn't have ever worked?  Perhaps it wasnt working lol.  I tested this and it was working.

MattMcFarland:
Ok I was wrong..

@YerkoPalma I actually fixed a bug :)

vanilla example only works with the changes in this PR :)

YerkoPalma:
Yeah, I noticed when I try to make a choo example on jsfiddle. thats why I didn't understand the idea of separate things :smile: I think that the PR should be merged and the html package, that is just yo-yo, should keep the same :+1:

MattMcFarland:
Can we close this one? :)

PR #208, opened 2016-08-02
Fixes for #160

MattMcFarland:
Per #160 I updated the build to add a separate html.js file.

I was thinking though, if we separate them, shouldn't we exclude html from the main choo.js umd build?

Also I had to merge / rebase to update my fork which created two commits :(  I'm not sure how to squash them though.


yoshuawuyts:
@MattMcFarland sweet! - so we're already stripping `html` from the main UMD build; `yo-yoify` takes care of that. Let's see if I can rebase your PR correctly and get it in here. Thanks! :sparkles:

MattMcFarland:
I'm not sure what happened I tried to squash commits :/

MattMcFarland:
Ok here is the proper PR: #209

PR #209, opened 2016-08-02
Fixes for #160

MattMcFarland:
- Update build script to create the html.js and html.min.js file as requested per the issue.
 - Added the html.min.js file to the vanilla example
 - Tested and working as expected.


timwis:
👍 LGTM, thanks!

yoshuawuyts:
Sweet, thanks!

PR #214, opened 2016-08-07
location-model: split to separate file

yoshuawuyts:
Split out the location model into a separate file. Feels like a more natural separation C:


yoshuawuyts:
nah

Issue #166, opened 2016-07-13
I want to refetch data when i switch route, is the subscriptions' right place to work with?


crapthings:
what is best practical or preferred way to get inital data when switch route ? 

users.js

```
require('./model')

module.exports = (state, prev, send) => render`
  <div onload=${send('fetch')}> // i've got infinite render loop
    <h3>user list</h3>
    ${userForm(state, prev, send)}
    ${userList(state, prev, send)}
  </div>
`

function userList(state, prev, send) {
  return render`
    <table>
      <thead>
        <tr>
          <th width='60'>avatar</th>
          <th width='200'>name</th>
          <th width='200'>username</th>
          <th width='300'>email</th>
          <th width='200'>website</th>
          <th>phone</th>
          <th>lorem</th>
        </tr>
      </thead>
      <tbody>
        ${state.users.map((user) => render`
          <tr>
            <td>
              <img src='${user.avatar}' alt='${user.name}' class='img-avatar' />
            </td>
            <td>${user.name}</td>
            <td>${user.username}</td>
            <td>${user.email}</td>
            <td>${user.website}</td>
            <td>${user.phone}</td>
            <td>${user.lorem}</td>
          </tr>
        `)}
      </tbody>
    </table>
  `
}

function userForm(state, prev, send) {
  return render`
    <form onsubmit=${submit}>
      <input type='text' id='username' placeholder='enter an username' required autofocus />
      <input type='submit' value='submit' />
      <input type='button' value='refetch' onclick=${refetch} />
    </form>
  `

  function submit (e) {
    e.preventDefault()
    let $username = document.getElementById('username')
    send('add', { payload: { username: username.value } })
    $username.value = ''
  }

  function refetch() {
    send('fetch')
  }
}
```

model.js

```
const faker = require('faker')

const _ = require('lodash')

app.model({

  state: {
    users: []
  },

  subscriptions: [
    (send, done) => send('fetch', done)
  ], // subscriptions only happen when the page reload, not on change route

  effects: {
    fetch: (data, state, send, done) => {
      let users = _.times(10, n => {
        return _.extend(faker.helpers.userCard(), {
          avatar: faker.internet.avatar(),
          lorem: faker.lorem.sentence()
        })
      })

      send('init', { payload: users }, done)
    }
  },

  reducers: {
    init: (data, state) => {
      return {
        users: data.payload
      }
    },

    add: (data, state) => {
      return {
        users: state.users.concat(data.payload)
      }
    }
  }

[...]

PR #227, opened 2016-08-20
prev.params should always exist

timwis:
The code I suggested in [#166 (comment)](https://github.com/yoshuawuyts/choo/issues/166#issuecomment-241204080) requires you check if `prev.params` exists before checking one of its properties, otherwise you'll throw a runtime error on the first load.

```javascript
const view = (state, prev, send) => {
  if (!prev.params || !prev.params.user || prev.params.user !== state.params.user) {
    send('fetch', state.params.user)
  }
  return html`<div>...</div>`
}
```

This pull request lets you always check properties of `prev.params`:
```javascript
const view = (state, prev, send) => {
  if (!prev.params.user || prev.params.user !== state.params.user) {
    send('fetch', state.params.user)
  }
  return html`<div>...</div>`
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -119,7 +119,7 @@ function choo (opts) {
   // create a new router with a custom `createRoute()` function
   // (str?, obj, fn?) -> null
   function createRouter (defaultRoute, routes, createSend) {
-    var prev = {}
+    var prev = { params: {} }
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
:100: yeah nice one

yoshuawuyts:
Wooh!

PR #237, opened 2016-08-27
Fixing consistency of signatures.

pekeler:
`data` instead of `action`. Same as https://github.com/trainyard/choo-cli/issues/20 but also including effects and lifecycle methods.


yoshuawuyts:
Thanks!

Issue #232, opened 2016-08-26
AppStart hook


YerkoPalma:
Is there any `startup` hook or something like that on choo? I'm asking thinking in plugin development, I know that I could just use a content loaded event or something, but I would like to write some plugin that fire an event on startup, is there any hook for that? if it isn't, can I use one of the availaible hooks for something like that?

Thanks

yoshuawuyts:
Yeah there is none; `choo` is loaded when the dom is done loading so it's best to create a separate package for that. If you wanna do something dirty you can write a function inside a `subscription` that will be called. There's no hard guarantees when `subscriptions` are done loading though, so YMMV. Does this answer your question?

YerkoPalma:
Yeah, that's what I though. The thing is, that plugins only define hooks, not subscriptions. Is not actually hard to write some function called on the dom load event, it's just not _the choo way_ IMO.
Thanks for the quick answer.

yoshuawuyts:
no yeah you're right - perhaps we should add one. Curious what the use case is though

YerkoPalma:
In my case, I'm playing with service workers, there is an initial step of registering, that has to be done on app start, is pretty similar for almost every case, so is like copy and paste always. Is so mechanical that I though is perfect for a plugin that provides offline support automagically :+1:

yoshuawuyts:
Ahh, yeah that's a good use case. I'm cool with adding an extra hook for this hey

YerkoPalma:
Just curious, this hook should be added on choo or barracks?

yoshuawuyts:
I think `choo` as `barracks` things could be started at different times, making it a weird abstraction - the thing calling `barracks` will know when it's all booted up so that'd be the place

PR #240, opened 2016-08-29
added onStart hook

YerkoPalma:
ref: #232 

This probably isn't ready to merge yet, because:

- There is no test (I couldn't figure out _how_ to add them).
- It doesn't work as expected. It's called before the app is added to the DOM and has no access to the state.

The thing of having it before the DOM is updated is not that important, but it definitely should have access to the state. Any help or opinion is appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,6 +497,9 @@ There are several `hooks` that are picked up by `choo`:
   `action` is fired.
 - __onStateChange(data, state, prev, caller, createSend):__ called after a
   reducer changes the `state`.
+- __onStart():__ called on `choo.start` _before_ the app is appended to the DOM,
+  and also with no access to the model state. This hook is intended for
+  initialize of some third party libraries.
 
 __:warning: Warning :warning:: plugins should only be used as a last resort.
 It creates peer dependencies which makes upgrading versions and switching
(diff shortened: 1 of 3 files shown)


timwis:
Hey @YerkoPalma, I think I recall this being mentioned in IRC. Remind me / those not there of the use case?

YerkoPalma:
The use case is basically initialization of librarys or APIs, like service workers, indexedDB or something other. Now, the `onload` event of bel is definitely a good choice too, but the thing is those events are not set on the `choo.use` function, so they can't be wrapped in a plugin. 

If those events could be wrapped in the use function to be used as a plugin it would be awesome, and this PR would have no sense :smile:

ajoslin:
Wouldn't it make sense to do this sort of initialization in your own function around choo.start?

```js
function startMyApp () {
  myService.init()
  return app.start()
}
```

YerkoPalma:
@ajoslin But that wraping function has the same problems described above, besides can't be used in a plugin. The main intention of this PR is to be able to do this

```javascript
const startPlugin = require('start-plugin')
const choo = require('choo')

choo.use(startPlugin())

choo.start()
// ...
```

YerkoPalma:
closes in favor of #249

PR #242, opened 2016-09-01
Add example using custom elements

savelbr:
Inspired by [this article](https://medium.com/dev-channel/the-case-for-custom-elements-part-1-65d807b4b439), I decided to try and use a basic Custom Element within `choo`. Feel free to merge if you think this could be useful as an example :)

### DIFF
--- a/examples/custom-elements/client.js
+++ b/examples/custom-elements/client.js
@@ -1,0 +1,35 @@
+const choo = require('../..')
+const html = require('../../html')
+
+const app = choo()
+
+app.model({
+  state: {twelveHour: false},
+  reducers: {
+    'toggle': (data, state) => ({ twelveHour: data })
+  }
+})
+
+const elements = [
+  {name: 'live-clock', proto: require('./clock')}
+]
+
+const view = (state, prev, send) => {
+  return html`
+    <main>
+      <h1>
+        <live-clock data-twelve=${state.twelveHour}></live-clock>
+      </h1>
+      <button onclick=${() => send('toggle', !state.twelveHour)}>
+        Toggle ${state.twelveHour ? '24h' : '12h'}
+      </button>
+    </main>
+  `
+}
+
+app.router('/', route => [
+  route('/', view)
+])
+
+elements.forEach(element => document.registerElement(element.name, element.proto))
+document.body.appendChild(app.start())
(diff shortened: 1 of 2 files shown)


savelbr:
I'm personally not too keen on the `class` syntax, but I've yet to come across a cleaner way to define Custom Elements. Feedback welcome :)

yoshuawuyts:
Heya, thanks for the great work! - I like it when people come up with stuff like this. I'm a bit on the fence about this PR; so let me try and explain what I'm thinking here.

Right now the `examples/` directory in `choo` is a bit of a smoke test. When a new version with breaking changes comes out, it's the first thing we check to see if the API is right. We test this for all the examples. Thus all examples that are in that directory should have a unique value proposition regarding `choo`. I feel that while this is a great example of how to use `custom elements` with `choo`, the thing that makes it special is less technical and more like reference material. Perhaps this would be best as a separate repo we could link to from [awesome-choo](https://github.com/YerkoPalma/awesome-choo)?

And now for the more opinionated part: I've had a bit of a history with webcomponents, and am not entirely convinced of their added value. To me declaring globals that propagate throughout the whole application code feels like something that should preferably be avoided. With [bel](https://github.com/requireio/custom-element) making its appearance, we've now found a way to create DOM-native elements that can persist their own state between renders, and can be rendered on both the client and the server. It completely mitigates the need to have `webcomponents` in any realistic scenario.

Something that has piqued my interest though, is to find a way to sneak `bel` elements into `webcomponents`. A bit of a Trojan horse if you will. This could be done by wrapping `bel`'s lifecycle hooks into the equivalent webcomponent methods, so they could just be rendered as such. I'd be happy to build it, but given that I'm not too invested in webcomponents perhaps someone else would be interested in giving it a spin? The results would be good I reckon.

Hope this sheds a bit of light on my thinking here; I'm not sure if this is a coherent story or if it answers your question but I hope it's useful. Cheers! :sparkles:

savelbr:
Right, if given the choice I'd also write a `bel` element over a `webcomponent` any day. What interests me about `custom elements` though is that, being a web standard, they could a very effective tool to break down the existing walled gardens you talk about in the README. In that regard, your Trojan horse idea sounds especially interesting. When I have some time, I might give that a try in a separate repo.

PR #244, opened 2016-09-02
Fix link to browserify

gsf:


timwis:
@gsf! You made it! Thanks for the fix :)

PR #254, opened 2016-09-12
Emoji in list on github doesnt work anymore...

reminyborg:
The emojis works quite well as a "bullet" anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,23 +105,18 @@
 - __very cute:__ choo choo!
 
 ## Demos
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
- :cry: seems like a bug
- 😭 truly does
- 😭 have you tested if using the unicode directly does work?

reminyborg:
Will try the unicode.
It is sad. But sometimes i just like to flow around the bugs. 💧

reminyborg:
Updated this to use unicode. Seems to work nicely, thought might be better to fix in github/markdown.

Issue #204, opened 2016-07-31
[release] 4.0 tracker


yoshuawuyts:
## Major release, routing edition

Programmatic navigation is essential. But to achieve it we need to have breaking changes. And if we're going to break things, we might as well go all out again. So this issue is here to track the changes for the `4.0.0` release. All patches will be merged into [the 4 branch](https://github.com/yoshuawuyts/choo/tree/4), so that should provide an accurate representation of where things are at. Feedback and testing at any stage would be very welcome!
### Changes
- [x] slim down server side rendering API | [issue](https://github.com/yoshuawuyts/choo/issues/191) | [pull-request](https://github.com/yoshuawuyts/choo/pull/203)
- [x] update router API to be lisp-like | [issue]() | [pull-request]()
  - [x] update `sheet-router` to be lisp-like | [issue](https://github.com/yoshuawuyts/sheet-router/issues/29) | [pull-request](https://github.com/yoshuawuyts/sheet-router/pull/33)
  - [x] update `sheet-router` to add thunking | [issue](https://github.com/yoshuawuyts/sheet-router/issues/17) | [pull-request](https://github.com/yoshuawuyts/sheet-router/pull/34)
  - [x] update `wayfarer` to add a `walk` function | [issue](https://github.com/yoshuawuyts/wayfarer/issues/43) | [pull-request](https://github.com/yoshuawuyts/wayfarer/pull/45)
  - [x] update `sheet-router` to expose the `walk` function | [issue](https://github.com/yoshuawuyts/sheet-router/issues/30) | [pull-request](https://github.com/yoshuawuyts/sheet-router/pull/38)
  - [x] update `choo` to use the `walk` function | [pull-request](https://github.com/yoshuawuyts/choo/pull/206)
- [x] support inline anchor links | [issue](https://github.com/yoshuawuyts/choo/issues/65)
  - [ ] allow bypassing of link clicks in `sheet-router` | [issue](https://github.com/yoshuawuyts/sheet-router/issues/15) | [pull-request](https://github.com/yoshuawuyts/sheet-router/pull/27)
- [x] update router API to handle hashes by default
  - [x] update `sheet-router` to handle hashes by default | [issue](https://github.com/yoshuawuyts/sheet-router/issues/31)
- [x] update `location` state to expose `search` parameters (query strings) | [issue](https://github.com/yoshuawuyts/sheet-router/issues/31)
[...]

PR #268, opened 2016-10-07
feat(api:) arg order

### LINKED ISSUE #179: Reducer parameter order is reversed from how traditional reducers work, why? Should it be swapped to prevent confusion and make it easier to convert from Redux?
Array, Redux, and rxjs reducers all have the convention that the accumulated state is the first parameter. 

Why have you instead chosen to have the data/action to be first? That just seems like it is going to cause lots of confusion when people are used to using reducers elsewhere and now in this case it is backwards.

I know that I would be frustrated having to use one convention with arrays and then the opposite when working with choo. 

Also since many people might be coming to choo from React + Redux, all of their existing redux reducers will need to be reworked to use them in choo, otherwise the conversion would be much easier.

Would you consider swapping the order to use the common accepted signature (state, data)? or if not then is there a good reason?

yoshuawuyts:
Changes the argument order. Closes #179 

### Changes
- swap argument order :sparkles:
- replace last remaining `action` arguments with `data` for consistency
- update to `barracks@9`
- fixed a stubbed test (which was hanging)
- ref: #204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,14 @@ we support - while secretly sneaking in some performance upgrades. Enjoy!
 - [choo-handbook/upgrading/4.0.0]()
 
 ### changes
-- slim down server side rendering API |
+- :exclamation: slim down server side rendering API |
   [issue](https://github.com/yoshuawuyts/choo/issues/191) |
   [pull-request](https://github.com/yoshuawuyts/choo/pull/203)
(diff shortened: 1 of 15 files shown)


yoshuawuyts:
Merging as there's very few interesting things going on here; mostly a regex replace - not going to release v4 yet tho soooo can merrily merge and wait for humans to review-after-merge hah

PR #269, opened 2016-10-07
feat(http): remove

yoshuawuyts:
Removes `choo/http`. In previous IRC conversations we kind of came to the conclusion that this doesn't make a lot of sense to bundle in anymore given that a lot of choo's narrative is now about how to glue views, logic and state. We weren't mentioning it anywhere in the README anymore anyway, so it might make sense to deprecate in its entirety.

Having thoughts on this would be highly welcome! - Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ we support - while secretly sneaking in some performance upgrades. Enjoy!
   [pull-request]()
 - :exclamation: swap `state` and `data` argument order |
   [issue](https://github.com/yoshuawuyts/choo/issues/179)
+- :exclamation: remove `choo/http`. Use [xhr](https://github.com/naugtur/xhr)
+  instead | [pull-request](https://github.com/yoshuawuyts/choo/pull/269)
 - update `router` to use memoization |
   [issue](https://github.com/yoshuawuyts/sheet-router/issues/17) |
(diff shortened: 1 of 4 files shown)


tunnckoCore:
Yea. I thought for this too. Because I was playing with `choo` these days and its size browserify bundled isn't 7kb, nor 4kb. It was 30kb+.

```sh
% ❯ ls -al dist                                                                                           October 08, 03:03:49
total 384K
drwxr-xr-x 2 charlike users 4.0K Oct  5 05:12 ./
drwxr-xr-x 4 charlike users 4.0K Oct  5 05:17 ../
-rw-r--r-- 1 charlike users  152 Oct  5 05:17 200.html
-rw-r--r-- 1 charlike users 273K Oct  5 05:17 bundle.js
-rw-r--r-- 1 charlike users  38K Oct  5 05:17 bundle.min.js
-rw-r--r-- 1 charlike users  56K Oct  5 05:17 bundle.min.js.map
```
with `client.js` example from the README.

yoshuawuyts:
@tunnckoCore you might want to consider running gzip and other optimizations: `npm run instrument:gzip` will create an optimized bundle

yoshuawuyts:
everyone seems to be ok with this, merging!

PR #271, opened 2016-10-08
fix(href): fix routing

yoshuawuyts:
fixes bugs from the other routing patch


yoshuawuyts:
ohey look, bug fixes! :tada:

PR #279, opened 2016-10-13
fix(hash): make hash routing work

sethvincent:
This gets hash routing working based on the example I've been working on over here: https://github.com/sethvincent/examples-choo-v4/blob/master/hash-routes.js

Also adds the `onHash` listener as a subscription, as that seems appropriate for this use case.


index.js
```diff
@@ -73,7 +74,8 @@ function choo (opts) {
   function render (state, data, prev, name, createSend) {
     if (!_frame) {
       _frame = nanoraf(function (state, prev) {
-        const newTree = _router(state.location.pathname, state, prev)
```
yoshuawuyts: oh shit, you're right - I had the idea of passing in `state.location.href` and hash routing should be picked up by default now - could you perhaps give that a spin?

index.js
```diff
@@ -194,6 +196,14 @@ function createLocationModel (opts) {
       }
     }
 
+    if (opts.hash !== false) {
+      subs.handleHash = function (send, done) {
```
yoshuawuyts: I've been thinking about this: this only handles `window.hash = "#something"` API calls, and I'm not sure we should encourage that interface. If some component needs to change the route it can already do so by passing `send('location:set', '#something')`.

If for some reason something deeper down the tree produces a side effect using `window.hash =` I think we should consider that to be bad practice and encourage developers to not provide side effects like that. Does that make sense?

yoshuawuyts:
@sethvincent thanks heaps for chasing this up! Was literally missing the bandwidth to do it myself; this is great :grin:

index.js
yoshuawuyts: this would indeed also mean that we first check for `#something` on the page, and kinda cross our fingers duplicates don't exist in the router that would cause a route change. When using hashes for only one type of navigation this conflict should never occur, and if for some reason both types of routing need to exist the addition of the `state.location.query` parameter could provide an escape hatch for the routing (as GH does on their profile page). Reckon this is good enough? - if we go forward with this it would probably be worth documenting in detail haha

sethvincent: `state.location.href` throws this: `Uncaught Error: route ':9966/' did not match`

Would have to fix sheet-router to exclude the port, which is probably a reasonable thing to do anyway.

Yeah I feel like both types of routing shouldn't exist in one app. Does sheet-router need to change to recognize routing with `state.location.query`?

index.js
sethvincent: Oh for sure, that makes sense. I'll remove the `onHash` stuff.

yoshuawuyts:
oops, I updated the `4.0.0` branch to resolve merge conflicts; a simple rebase for this patch should work I reckon :sparkles:

sethvincent:
uggg, made a fresh start with #282 because morning/git/coffee/brain.

PR #289, opened 2016-10-23
feat(mount): copy {script,link} tags

yoshuawuyts:
Allows us to create root views that look like:
```js

function mainView (state, prev, send) {
  return html`
    <body>
      <main>
        <h1 class="f-6">choo choo</h1>
      </main>
    </body>
  `
}
```
And these will have like all the `<script>` and `<link>` tags copied over, without triggering again and other weird stuff :ok_hand:

### DIFF
--- a/mount.js
+++ b/mount.js
@@ -12,6 +12,17 @@ function mount (selector, tree) {
   documentReady(function onReady () {
     const oldTree = document.querySelector(selector)
     assert.ok(oldTree, 'could not query selector: ' + selector)
+
+    // copy script tags from the old tree to the new tree so
+    // we can pass a <body> element straight up
+    if (oldTree.nodeName === 'BODY') {
+      oldTree.childNodes.forEach(function (node) {
+        if (node.nodeName === 'SCRIPT') {
+          tree.appendChild(node.cloneNode())
+        }
+      })
+    }
+
     const newNode = yo.update(oldTree, tree)
     assert.equal(newNode, oldTree, 'choo/mount: The DOM node: \n' +
       newNode.outerHTML + '\n is not equal to \n' + oldTree.outerHTML +


timwis:
Hm, I've never seen a `<link>` tag in the `<body>`, only in the `<head>`. Maybe that tag's not necessary to check?

Would you still initiate choo via `const tree = app.start(); document.body.appendChild(tree)` ?

ungoldman:
why not `<html>`? :trollface:

yoshuawuyts:
lol, go away 👻

yoshuawuyts:
ok turns out some of my replies haven't come through. @timwis I replied:
> Haha, oops - yeah you're right :D

> Yeah nah, in this case you couldn't do that - appending body to body wouldn't work.
Right now `mount()` is being used for server side rendering, but it could be used
to replace any type of node - in this case we're going for the body, which I guess is
also already rendered on the server so it kinda fits in that line hah

Second email:
> Updated PR :sparkles:

yoshuawuyts:
margin!

yoshuawuyts:
lol, go away 👻

yoshuawuyts:
Updated PR :sparkles:

ungoldman:
hmm weird, looks like those email responses finally came through! I responded to another issue over email this morning and didn't see it show up. betting @github's message queue is borked somewhere

yoshuawuyts:
yeah super odd

yoshuawuyts:
Haha, oops - yeah you're right :D

Yeah nah, in this case you couldn't do that - appending body to body
wouldn't work.
Right now `mount()` is being used for server side rendering, but it could
be used
to replace any type of node - in this case we're going for the body, which
I guess is
also already rendered on the server so it kinda fits in that line hah

ungoldman:
👻  ghost in the machine

PR #293, opened 2016-10-25
Fix router example in README

aknuds1:
AFAICT `app.router` expects a list of pairs, whereas the README exemplifies with a single pair.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,9 @@ function mainView (state, prev, send) {
   `
 }
 
-app.router(['/', mainView])
+app.router([
+  ['/', mainView]
+])
 
 const tree = app.start()
 document.body.appendChild(tree)


yoshuawuyts:
oh yeah, no this works as expected (:

yoshuawuyts:
__edit:__ to clarify: the router now uses recursive pattern matching on arrays and just like understands the pattern so it works as expected :sparkles:

aknuds1:
@yoshuawuyts Ah yeah it does work actually, I must've done something wrong when I tested it before.

Maybe it makes more sense to use a nested array in the example though? I mean, who is going to invoke the router with just one route in practice?

yoshuawuyts:
Haha, yeah maybe - though in all examples we like actually just use a
single route soooo. I think the best way forward would be to create a
better story in the handbook perhaps; there's a lingering router section
waiting to be filled haha
https://github.com/yoshuawuyts/choo-handbook/tree/master/04_routing

PR #294, opened 2016-10-25
Preserve state.location.pathname

aknuds1:
Preserve state.location when calling route handler.


yoshuawuyts:
yeah so if this is not working it signals a deeper lying problem; it means the path resolution algorithm is getting out of sync - a repro where this is happening would very much be appreciated!

aknuds1:
@yoshuawuyts OK will try.

yoshuawuyts:
This should be solved in #343

PR #297, opened 2016-10-27
deps: use nanomorph

ym97:


yoshuawuyts:
Thanks for the PR - nanomorph is currently unstable so landing it doesn't make sense at the moment. Thanks!

PR #320, opened 2016-11-07
feat(docs): update for 4.0.0

yoshuawuyts:
IMPLEMENTS DOCUMENTATION FOR 4.0.0 LET ME KNOW IF I MISSED ANYTHING. ALSO SORRY FOR THE ALL CAPS COMMIT MESSAGE I TRULY AM

PS. HOPEFULLY THE WORD "ACTION{S}" NOW MAKES MORE SENSE AND BECOMES EASIER TO REASON ABOUT

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,22 +152,18 @@ const tree = app.start()
 document.body.appendChild(tree)
 ```
 
-To run it, save it as `client.js` and run with [budo][budo] and
-[es2020][es2020]. These tools are convenient but any [browserify][browserify]
-based tool should do:
+To run it, save it as `client.js` and run with [bankai][bankai]. `bankai` is
+convenient but any [browserify][browserify] based tool should do:
 ```sh
-$ budo client.js -p 8080 --open -- -t es2020
-```
+# run and reload on port 8080
+$ bankai client.js -p 8080 --open
 
-And to save the output to files so it can be deployed, open a new terminal and
-do:
-```bash
-$ mkdir -p 'dist/'
-$ curl 'localhost:8080' > 'dist/index.html'
-$ curl 'localhost:8080/client.js' > 'dist/client.js'
+# compile to static files in `./dist/`
+$ bankai build index.js dist/
+
+# deploy to github pages using `tschaub/gh-pages`
+$ gh-pages -d dist/
 ```
-All using a couple of shell commands and `.js` files, no grandiose boilerplate
-needed.
 
 ## Philosophy
 We believe programming should be fun and light, not stern and stressful. It's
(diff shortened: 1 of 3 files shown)


YerkoPalma:
I think you have missed a v4 badge, actually there is only a v3, but with the new version it would be really helpful to know which choo are we looking at.

marlun:
`choo-model` seem to be used without introduction?

yoshuawuyts:
ya it does; do you find that confusing?  - wasn't sure myself; but I like
the API a bit better tho it's higher level - should we maybs just stick to
the built-in one?

timwis:
Woo! One suggestion: The description of an effect doesn't really convey what they are. For people who are not coming from redux-land they may not be familiar with the concept.

> `effects` are used to do a thing, and call another `action` when the thing is done.

Perhaps we could be a bit more descriptive on what a "side effect" is? Here's the verbiage we use in [your first app](https://yoshuawuyts.gitbooks.io/choo/content/02_your_first_app.html) in case it's helpful:

> Effects are similar to reducers except instead of modifying the state they cause side effects by interacting servers, databases, DOM APIs, etc. Often they'll call a reducer when they're done to update the state. For instance, you may have an effect called getUsers that fetches a list of users from a server API using AJAX. Assuming the AJAX request completes successfully, the effect can pass off the list of users to a reducer called receiveUsers which simply updates the state with that list, separating the concerns of interacting with an API from updating the application's state.

README.md
```diff
@@ -375,21 +363,27 @@ from within namespaced `models`, and usage should preferably be kept to a
...
 
-Views are also passed the `send` function, which they can use to dispatch actions that can update the state. For example, the DOM tree can have an `onclick` handler that dispatches an `add` action.
+Views are also passed the `send` function, which they can use to dispatch
+actions that can update the state. For example, the DOM tree can have an
+`onclick` handler that dispatches an `add` action.
 
-```javascript
-const view = (state, prev, send) => {
+```js
+function view (state, prev, send) => {
```
timwis: forgot to remove `=>`

marlun:
> ya it does; do you find that confusing?  - wasn't sure myself; but I like
the API a bit better tho it's higher level - should we maybs just stick to
the built-in one?

I think we should stick to the built-in functionality unless we feel that everyone should use choo-model instead and then we should probably only use that in the code examples and not show the more low level one?

Looking at https://github.com/yoshuawuyts/choo-model I'm not sure what the benefits are? 😄

If we want to show them both in the README I at least believe we should have some kind of introduction to it before using it in an example.

marlun:
I've just read the "Concept" part a couple of times and feel like it could be kept a lot shorter and more to the point and let the rest of the README go more into specifics.

The following is not complete (I don't mention namespaces at all) or even good but maybe it at least shows what I mean :smile:

_choo cleanly structures internal data flow, so that all pieces of logic can be
combined into a nice, cohesive machine. Roughly speaking there are two parts
to choo: the views and the models. Models take care of state and logic, and
views are responsible for displaying the interface and responding to user
interactions._

_All of choos state is contained in a single object and whenever it changes the
views receive a new version of the state which they can use to safely render a
complete new representation of the DOM. The DOM is efficiently updated using
DOM diffing/patching._

_The logic in choo exist in three different kinds of actions, each with their
own role: effects, subscriptions and reducers._

  * _**Effects** makes an asynchronous operation and calls another action when it's
    done._
  * _**Subscriptions** (called once when the DOM loads) listens for external input
    like keyboard or WebSocket events and then calls another action._
  * _**Reducers** receives the current state and returns an updated version of the state which is then sent to the views._

 ```txt
 ┌─────────────────┐
 │  Subscriptions ─┤     User ───┐
 └─ Effects  ◀─────┤             ▼
 ┌─ Reducers ◀─────┴─────────── DOM ◀┐
 │                                   │
 └▶ Router ─────State ───▶ Views ────┘
```

Ryuno-Ki:
"calls another action" … so in theory a Subscription could call a subscription? Or an effect another effect?

bengourley:
I noticed the router example wasn't updated so I fixed it and made a PR into this branch: https://github.com/yoshuawuyts/choo/pull/337

yoshuawuyts:
oops sorry for getting stuck on this PR - been doing a lot of travel and work lately ;_; - thanks everyone for chiming in!

tunnckoCore:
Haha. `choo-model` looks just what I was thinking last night - more declarative approach.

@Ryuno-Ki yea.

yoshuawuyts:
coool, updated all the things - added `standard-markdown` to lint the code.

@marlun @timwis the writing you did was really good; if you've got time having more of it would be amazing! :D

tunnckoCore:
@yoshuawuyts fantastic. So will wait it these days :)

PR #337, opened 2016-11-22
docs: update router example in readme

bengourley:
Fix the router example for the updated interface in v4.


yoshuawuyts:
thanks a lot! :grin:

PR #343, opened 2016-11-28
fix(router): fix location on newstate

yoshuawuyts:
Fixes location being passed on a first render; we were incorrectly reading from a nonexisting previous render which resulted in the location being dropped in a first render - not good

### DIFF
--- a/index.js
+++ b/index.js
@@ -111,7 +111,7 @@ function choo (opts) {
   // create a new router with a custom `createRoute()` function
   // (str?, obj) -> null
   function createRouter (routerOpts, routes, createSend) {
-    var prev = {}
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
oh god this patch was wrong; updating! - the revamped `sheet-router/create-href` method takes care of the issues we used to have (e.g. parts of the destructured route state not being updated) so doing it inside `choo` no longer makes sense. In this case reading the old value was causing a bug where route updates were one tick behind causing super weird things to happen.

Glad I ran into this before merging and not someone else at a later stage hah

yoshuawuyts:
ya ok merging :sparkles:

PR #347, opened 2016-12-05
Added example app: Thing Translator

dmotz:
Been building a few projects with Choo and it's been great!

Recently open-sourced one of them and thought it might serve as a good example app since it's fairly tiny (https://github.com/dmotz/thing-translator).

Thanks for Choo!


tunnckoCore:
looks like an amazing idea! it has some bugs when i tried it. for example can exit from it so i fastly should remove url bar when unlock the screen, dont know

timwis:
Wow, that's a cool app. You should consider putting it in [awesome-choo](https://github.com/YerkoPalma/awesome-choo)

yoshuawuyts:
omg this is the coolest!

PR #350, opened 2016-12-10
prev.location should always exist

timwis:
Same logic as #227

### DIFF
--- a/index.js
+++ b/index.js
@@ -119,7 +119,7 @@ function choo (opts) {
   // create a new router with a custom `createRoute()` function
   // (str?, obj, fn?) -> null
   function createRouter (defaultRoute, routes, createSend) {
-    var prev = { params: {} }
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
In 4.x `prev` will be `null` if it doesn't exist in a prior run - it turned out that having `prev.location` created way more complicated checks; e.g.
```js
// this is the case in v4 now:
if (prev) // do thing with prev.location.search

// instead if prev.location always exists:
if (prev && prev.location && prev.location.search) // do thing with prev.location.search
```

In v4 I feel we're closer to being honest by allowing `prev` to `null` on a first render and then always have a destructured `prev.location` setup.

Does this make sense?

timwis:
Hm, @yoshuawuyts your requested change isn't showing up for me - it just says `_`

PR #352, opened 2016-12-11
[major] choo v4

yoshuawuyts:
Update choo to version 4 :sparkles: - changes are included in `CHANGELOG.md`. I'd copy and paste it here, but my clipboard has crashed and I'm live streaming so lol can't reboot computer


examples/mailbox/views/mailbox.js
```diff
@@ -7,11 +7,16 @@ const nav = require('../elements/nav')
...
       ${pathname(state, prev, send)}
       ${nav(state, prev, send)}
       <section class="fl mt4 w-80 db">
         ${emailList(state, prev, send)}
       </section>
     </main>
   `
+
+  function goHome () {
+    send('location:set', { pathname: '/' })
```
jessehattabaugh: is this supposed to be `{href: '/'}` ?

ebuckthal: This does require a bit of clarification. A quick demo shows that `{ href: '/' }` changes `window.location` and reroutes the way I expect it should. However, the documentation was updated to use `{ pathname: '/'}` which doesn't update the `window.location`, nor does it re-route correctly (though it does appear to change the state).

yoshuawuyts:
Yeah it should probably href as that updates the whole url - PR welcome for
docs fixes hah 😅

PR #355, opened 2016-12-11
Fix prev.location.params bug

### LINKED ISSUE #354: prev.location.params not being set
### Expected behavior
After the first tick (when `prev` would be `null`), `prev.location.params` should be an object containing the previous `params`.

### Actual behavior
`prev.location.params` is always an empty object

```javascript
const choo = require('choo')
const html = require('choo/html')

const app = choo()

app.model({
  state: { world: 'earth' },
  reducers: {
    updateWorld: (state, data) => ({ world: data })
  }
})

const View = (state, prev, send) => {
  console.log(state, prev)
  return html`
    <main>
      Hello, ${state.world}
      <button onclick=${onClick}>Update world</button>
    </main>
  `
  function onClick (e) {
    send('updateWorld', 'mars')
  }
}

app.router(['/:name', View])

const tree = app.start()
document.body.appendChild(tree)
```

screen shot 2016-12-11 at 15 15 21

^ the object on the left is `state`, the object on the right is `prev`. Notice that on the right, `params` is an empty object, despite there being a param in the `href` above it.

timwis:
Closes #354. Unless you can imagine this causing any regressions?

### DIFF
--- a/index.js
+++ b/index.js
@@ -126,12 +126,12 @@ function choo (opts) {
       const send = createSend('view: ' + route, true)
       return function chooWrap (params) {
         return function (state) {
-          const nwPrev = prev
-          prev = state
-
           // TODO(yw): find a way to wrap handlers so params shows up in state
           const nwState = xtend(state)
           nwState.location = xtend(nwState.location, { params: params })
+
+          const nwPrev = prev
+          prev = nwState // save for next time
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
`v4.0.1`

PR #359, opened 2016-12-12
feat(stop): expose barracks stop() fn

bengourley:
Bump barracks version and expose `store.stop()` on choo.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,6 +31,7 @@ function choo (opts) {
   start.router = router
   start.model = model
   start.start = start
+  start.stop = _store.stop
   start.use = use
 
   return start
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "barracks": "^9.0.0",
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -31,6 +31,7 @@ function choo (opts) {
   start.router = router
   start.model = model
   start.start = start
+  start.stop = _store.stop
```
yoshuawuyts: hmmm, I feel we might want to add this to a plugin somehow - make it hard to access. does that make sense? - or is that a silly thought?

bengourley: sure, I've just been using like this in my fork because it was super simple and it works! Did you have any particular api/implementation in mind?

bengourley: yo @yoshuawuyts, any thoughts on this?

yoshuawuyts:
`v4.1.0` - sorry for the delay on this! :grin: - should be all good now!

bengourley:
Woo! No worries, I've just been running my fork but can switch over now.
Thanks for merging 👌

PR #362, opened 2016-12-13
Update README.md with correct 'location:set' usage

reminyborg:
Struggled with this when upgrading to v4.0.


yoshuawuyts:
thanks!

PR #370, opened 2016-12-18
Update README.md with correct order of arguments in reducer

paulsonnentag:
I think I found an error in the readme where the order of arguments is still reflecting the old v3.


yoshuawuyts:
awesome, thanks!

PR #371, opened 2016-12-24
Update README.md: change parameter name in docu of plugin hooks

paulsonnentag:
Just a minor change in the README. I think in the documentation of the plugin hooks the parameter name of wrapInitialState should be named state instead of fn.


yoshuawuyts:
This actually takes a callback that provides the state, like so:
```js
app.use({
  wrapInitialState: function (state) {
    return /* do stuff with state here */
  }
})
```

PR #373, opened 2016-12-30
add a hook to wrap the router

yoshuawuyts:
Allows us to wrap the router, which is useful for stuff like fixing iframes in v4 (https://github.com/yoshuawuyts/choo/issues/344) and perhaps electron stuff also

Definitely needs tests to be added; a PR onto this branch for that would be :sparkles: amazing :sparkles: - also code reviews welcome! Thanks! :D

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 const createLocation = require('sheet-router/create-location')
 const onHistoryChange = require('sheet-router/history')
+const wrapHook = require('barracks/wrap-hook')
 const sheetRouter = require('sheet-router')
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
This would be a minor release - oh yeah also forgot to mention this would restore requirebin.com support - was struggling with that earlier - code to solve that is to iterate over the array of routes ahead of time and prepend them with the current url inside the iframe

aknuds1:
How about including a test, would help in understanding how this works?

PR #375, opened 2017-01-06
Update README to reflect current size

antiBaconMachine:
I really like this framework, great job.

However, I showed it to some colleagues and immediately they called bs on `5kb` framework size and well technically they are right. It's `7.3kb` plus a basically essential `11kb` html module for total of `18.3kb`. That's hardly something to be ashamed off, in fact it's great, but it's not `5kb` as stated.

It's a very minor niggle but hey, we all know how this industry attracts pedants and this framework lost a bit of credibility almost immediately because of this claim.

Sorry to be "that guy", please consider updating README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
   <strong>Fun functional programming</strong>
 </div>
 <div align="center">
-  A <code>5kb</code> framework for creating sturdy frontend applications
+  A <code>7kb</code> framework for creating sturdy frontend applications
 </div>
 
 <br />
@@ -94,7 +94,7 @@
 </details>
 
 ## Features
-- __minimal size:__ weighing `5kb`, `choo` is a tiny little framework
+- __minimal size:__ weighing `7kb` (`18kb` with optional html module), `choo` is a tiny little framework
 - __single state:__ immutable single state helps reason about changes
 - __small api:__ with only 6 methods, there's not a lot to learn
 - __minimal tooling:__ built for the cutting edge `browserify` compiler


timwis:
If I recall correctly, Yosh measured how much choo represents in a bundled app. And when you bundle it, you'd use unassertify and yo-yoify which significantly reduce the bundle size, not to mention uglify/uglifyify/gzip. You sound be able to reproduce it using a script in the scripts folder. But maybe things have changed with v4, I'll defer to @yoshuawuyts there.

antiBaconMachine:
The way I measured was to run `build:min` which does run both of those transforms and uglify, then I ran the two generated artefacts through gzip so unless there is some other magic involved I think my measurements stand.

Should probably have clarified in the original description so to be clear  

**choo is currently `18.3kb` minified and gzipped**

yoshuawuyts:
looks like we might have a regression when bundling, gonna dig in and see what's up. If it's anything it's probably browserify's `process` package being pulled in or something similar (:

yoshuawuyts:
Oh, also it's worth noting that the `html` package should not be included in production; as we mention in the readme for production releases you should strip the runtime compiler with static `document.createElement()` calls using [yo-yoify](https://github.com/shama/yo-yoify) - this both lowers the resulting bundle size and significantly improves performance.

So yeah, you shouldn't be hitting `18kb` in production, like ever

yoshuawuyts:
Ah, alright - yeah we made a bit of a regression in terms of size for the 4.x patch - just ran it again on my machine and everything's as expected. It comes in at `6.41kb` gzipped. The difference is probably because I'm running `zopfli` rather than stock gzip which on average saves somewhere around 20% I believe.

screen shot 2017-01-06 at 16 06 42

---

For some historical backdrop on why we're past 6kb now - in the 1.x days we were at ~6kb, we dropped to 4.6kb in 2.x because of `yo-yoify` - but then the addition of our plugin system, more advanced router and event system pushed us back past 6kb again.

We're probably going to shed like another 1-2kb if we'd move `on-load` out of core (currently included in `bel`) and replace `morphdom` with `nanomorph`. I suspect this might happen in the near future and I don't want to update the file size in the readme every time we patch a dependency in the tree - I'd rather spend my time differently.

---

__On a completely different note:__ this issue is actually getting to me. The tone of voice, especially the _Sorry to be "that guy"_ remark - it puts me on edge. I get that you wanna be straight forward, and think this helps - but to me it comes off as rude. You highlight one of your findings as fact, and make me feel like I need to reply immediately or else this could spin out of control. I'm not a fan of people dropping into my mentions like that. I'm sure you have no bad intent, but still I'd urge you going forward to consider the humans behind the project and exercising patience before making bold claims. Thanks heaps!

antiBaconMachine:
Oh I'm sorry, that was not my intention at all, I wasn't going to raise this at all but then I figured this was literally stopping someone else from looking at this framework which I really like and if it was mine I wouldn't want something so simple to do that.

However I had totally missed the point that the html stuff is supposed to be rewritten anyway. That's really interesting, I will look at it some more. If so that renders this more or less irrelevant feel free to close, sorry to waste your time 👍

yoshuawuyts:
@antiBaconMachine all good!

PS I just ran a lil test of what would happen if we adopted `nanomorph` today and we'd be back at `5.1kb` :tada:

screen shot 2017-01-06 at 17 07 58

The biggest dependency then becomes `sheet-router` which I'm quite sure we can optimize (:

screen shot 2017-01-06 at 17 14 17

yoshuawuyts:
Think this is quite enough for now tho (: Thanks for opening the issue; we'll keep the file size in mind going forward - perhaps we should also publish the gzip version to the CDN? Anyway, closing :sparkles:

PR #377, opened 2017-01-07
Fix state.location.params docs

yoshuawuyts:


yoshuawuyts:
just a docs change, merging!

PR #379, opened 2017-01-10
Depend on morphdom <2.2 due to regressions in later versions

aknuds1:
I experience at least one regression with morphdom >= 2.2, where `selectedIndex` doesn't get set on `select` elements. This doesn't happen with version 2.1. This change makes sure that choo locks down the version of morphdom. I'm not sure if it's the right way to do it since morphdom is an indirect dependency, so would value any input on how it eventually ought to be done.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "nanoraf": "^2.1.1",
     "sheet-router": "^4.0.1",
     "xtend": "^4.0.1",
-    "yo-yo": "^1.2.2"
+    "yo-yo": "^1.2.2",
+    "morphdom": "<2.2"
   },
   "devDependencies": {
     "append-child": "~1.0.0",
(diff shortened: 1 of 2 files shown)


aknuds1:
Ah, case in point, the test fails. Is it OK to tell dependency-check to ignore morphdom?

yoshuawuyts:
@aknuds1 we should patch `yo-yo` for this to work; could you perhaps PR it there ?

aknuds1:
PR-ed yo-yo.

Ryuno-Ki:
@aknuds1 Could you set a reference (aka link) to that PR?

aknuds1:
[yo-yo PR](https://github.com/maxogden/yo-yo/pull/61) @Ryuno-Ki

PR #385, opened 2017-01-15
Save url params in state

rahatarmanahmed:
```
[11:21:33] <yoshuawuyts> uehhh
```

This is the simplest way to make this work, but it mutates the state right before rendering. I'm not 💯 sure this is safe, but all other solutions I've thought of required breaking changes and rearranging a number of things.

I'm mostly just seeing if the tests pass on this b/c I can't get them to run on my local machine.

### DIFF
--- a/index.js
+++ b/index.js
@@ -128,9 +128,8 @@ function choo (opts) {
       const send = createSend('view: ' + route, true)
       return function chooWrap (params) {
         return function (state) {
-          // TODO(yw): find a way to wrap handlers so params shows up in state
+          if (state.location) state.location.params = params
           const nwState = xtend(state)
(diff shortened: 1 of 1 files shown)


rahatarmanahmed:
Oh, figured out that barracks v9.2.1 was causing the failures. With v9.2.0 at least the tests pass locally but I still don't feel so hot about this. Any better ideas are welcome!

yoshuawuyts:
yeah this is pretty gnarly b/c of the mutation stuff - feel like it could cause some races. Should def think of a better way to handle this hmmm - if anything we're overhauling tests slightly as per #test, we should think of testing this behavior hey

rahatarmanahmed:
I'll go ahead and close this then.

PR #397, opened 2017-01-26
Update README reflecting API changes

### LINKED ISSUE #396: Confusing docs on passing sheet-router opts
The [current docs](https://github.com/yoshuawuyts/choo#tree--appstartopts) states that it is `app.start()` that needs to be passed an opts object containing router config:
```javascript
app.start({ history: false, href: false });
```

However, currently these needs to be passed to the "constructor" function to work:
```javascript
var app = choo({ history: false, href: false });
```

ruudud:
Fixes #396.

Reading `index.js` I couldn't find any references to `opts.hash`. Is it still active?

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,6 +449,18 @@ the first time, consider reading through the [handbook] or
 Initialize a new `choo` app. Takes an optional object of handlers that is
 passed to [app.use()](#appusehooks).
 
+Opts can also contain the following values:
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Oh nice, thanks! - yeah `opts.hash` was deprecated in `v4` - we now handle hash routes out of the box like you would expect :sparkles:

ruudud:
So..new PR with `opts.hash` removed or will you fix? :smiley:

PR #408, opened 2017-02-07
Fix tests

yoshuawuyts:
Slight garbage patch. Doesn't actually fix our tests but narrows it down to a race condition (e.g. now sometimes fails).

- Removes a whole bunch of example code we weren't really using anyway. We should probs have like a single big example somewhere else I think
- Strips stuff from readme (e.g. super outdated browser test cov code)
- Replaces `const` with `var`

Thanks!


yoshuawuyts:
lol, tests still failing. ueh; race conditions

PR #421, opened 2017-03-02
Delay triggering of subscriptions

### LINKED ISSUE #420: Subscriptions may get invoked before router is created
Model subscriptions may get invoked by Choo at a stage where Choo hasn't yet created its router, as a result of [calling _store.start](https://github.com/yoshuawuyts/choo/blob/master/index.js#L65). This can lead to havoc in that subscriptions may trigger rendering, which again depends on the internal router reference.  When the internal router is null in this case, a bug occurs.

Is this down to a bug in Choo, i.e. that subscriptions might get invoked before the router is instantiated? Seems that way to me, but I can't tell for sure.

aknuds1:
Delay triggering of subscriptions until infrastructure (such as router) is initialized. Fixes #420.

**Please note that this change isn't well tested yet, ideally I would add an automated test as part of this PR before it gets merged. Feedback on my approach would be great.**

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ const nanoraf = require('nanoraf')
 const assert = require('assert')
 const xtend = require('xtend')
 const yo = require('yo-yo')
+const tick = require('nanotick')()
 
 module.exports = choo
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -62,9 +63,16 @@ function choo (opts) {
   // (str?, obj?) -> DOMNode
   function start () {
     _store.model(createLocationModel(opts))
-    const createSend = _store.start(opts)
+    const createSend = _store.start(xtend(opts, {
+      subscriptions: false,
+      noSubscriptions: true
```
aknuds1: Is it a design omission that there are both `subscriptions` and `noSubscriptions` options? Seems to me these two options overlap! Basically, you can't set just `subscriptions` as it will make Barracks think that subscriptions have been called, which isn't the case.

yoshuawuyts: oops, yeah that's an artifact: `noSubscriptions` was deprecated & should not be used

aknuds1: @yoshuawuyts See my [Barracks PR](https://github.com/yoshuawuyts/barracks/pull/92).

yoshuawuyts:
Closing because choo 5 happened - thanks heaps tho!

PR #425, opened 2017-03-13
implement choo v5 API

yoshuawuyts:
heaps of changes!

Mostly following what's in https://medium.com/@yoshuawuyts/choo-v5-bc775b007b5e, but we're not including the `component` thing, and we changed the router API slightly. All for more simplicity.

All the `lib/` stuff should become their own modules, but yeah it's late and I can't be bothered. Also lol, what are tests even.

This version is gonna be released as `v5.0.0-2` so y'all can have a play with it. Check out the `example.js` file for a full example. Probably wanna symlink in `nanomorph` until https://github.com/yoshuawuyts/nanomorph/pull/47 has been resolved.

Anyway, hope this is somewhat shows the direction I'm feeling for v5; feedback is heaps welcome!

__edit:__ also yay for less API surface - we also dropped down to 4.3kB which is like neat. Huzzah


abradley2:
:tada: 
Awesome stuff Yosh! Great to see Choo getting even better!

Ok, so it looks like with the change to mutable state, we no longer have the view signature of
```
function (state, prev, send) {
```

since prev state is not _a thing_ right? [I had one big use case for `prev`](https://github.com/abradley2/constituency/blob/master/client/pages/member-profile.js#L48)

Sometimes I have a view that displays in response to a route like `/posts/123`

So on view create, I'd add an xhr request in response to the `onload` hook, which grabs the "post". However, if the user navigates from `/posts/1` to `/posts/2`, the view isn't really created a second time. So `onload` doesn't work here. I need to grab my data when a piece of state has changed _and_ the view has mounted.

To sum up what I was doing to solve this- in the render function I would compare the current and previous parts of the `location.params` to see if the user had navigated- to the same page, but with different parameters that would warrant me to fire my the functions i normally call in the `onload` hook.

Some prior art to solving this issue:

In vue, the vue router has a `beforeRouteEnter` and (added in a later version) a `beforeRouteUpdate` hook that [will fire whenever the route is the same (and therfore won't switch components) but relevant route params have changed](http://router.vuejs.org/en/advanced/navigation-guards.html#incomponent-guards)

----

So my first piece of feedback-
what do you think about an `onupdate` hook to complement `onload`?

aknuds1:
@abradley2 You might want to check out my [choo-routehandler](https://github.com/aknuds1/choo-routehandler) complementary route handling framework for Choo. It lets you define loading hooks for routes, which is what you need right? Bear in mind it's written for Choo v4 though...

bcomnes:
My only suggestion would be to include a [nanocomponant](https://github.com/yoshuawuyts/nanocomponent) in the example, demonstrating the isSameNode method.

example.js
```diff
@@ -0,0 +1,353 @@
+var mutate = require('xtend/mutable')
+var expose = require('choo-expose')
```
timwis: What's this library? It's not on npm

example.js
```diff
@@ -0,0 +1,353 @@
...
+var css = require('sheetify')
+var html = require('bel')
+var choo = require('./')
+
+css('todomvc-common/base.css')
+css('todomvc-app-css/index.css')
+
+if (module.parent) {
+  exports.todoStore = todoStore
```
timwis: Might this be overkill for an example app? Or is there some reason to include this?

example.js
```diff
@@ -0,0 +1,353 @@
...
+    app.use(persist())
+    app.use(logger())
+  }
+  app.use(expose())
+  app.use(todoStore())
+
+  app.route('/', mainView)
+  app.route('#active', mainView)
+  app.route('#completed', mainView)
+  app.mount('body')
```
timwis: Is there no `app.start()` anymore? `index.js` suggests it still exists.

example.js
```diff
@@ -0,0 +1,353 @@
...
+    <body>
+      <section class="todoapp">
+        ${Header(state, emit)}
+        ${TodoList(state, emit)}
+        ${Footer(state, emit)}
+      </section>
+      <footer class="info">
+        <p>Double-click to edit a todo</p>
+        <p>choo by <a href="https://yoshuawuyts.com/">Yoshua Wuyts</a></p>
+        <p>Created by <a href="http://shuheikagawa.com">Shuhei Kagawa</a></p>
```
timwis: FYI, at first read, I thought this person co-authored choo. My guess is that they wrote the original version of the choo todomvc app.

example.js
```diff
@@ -0,0 +1,353 @@
...
+      </footer>
+    </body>
+  `
+}
+
+function todoStore () {
+  return function (state, emitter) {
```
timwis: At first read, this made me think models needed to return functions, which is a bit less..simple, but the readme suggests they don't need to. If the example is meant to illustrate the simplicity, perhaps it's better if this doesn't?

example.js
```diff
@@ -0,0 +1,353 @@
...
+    var localState = state.todos
```
timwis: What I'm gathering from this block of `localState` modifications is that the way you update state in choo `v5` is by modifying the global `state` object, which is passed down to "models" (things that are `use`d). That pattern is a bit unclear here because it _looks_ like we're modifying a local variable, and not actually persisting that globally anywhere. In fact, because of the way JavaScript variables work by reference with objects, we _are_ modifying a global variable, but you'd have to be familiar with that to pick up on what's going on here. My suggestion would be to make modifications to `state.todos.active = []` etc. directly for simplicity's sake.

example.js
```diff
@@ -0,0 +1,353 @@
...
+      localState = state.todos = {}
+
+      localState.active = []
+      localState.done = []
+      localState.all = []
+
+      localState.idCounter = 0
+    }
+
+    emitter.on('DOMContentLoaded', function () {
```
timwis: Why would we want to wait until the DOM is finished rendering to register our listeners and such?

yoshuawuyts:
@abradley2 The `prev` argument in choo 4 was intended for component-like systems. With the latest release of `nanocomponent` it kinda showed that the `prev` argument lead to more complex APIs, complicated internals and could leak if not used right.

We also removed our internal copy of the `location` stuff because keeping it in sync was trouble, and if you want access to the location there's `window.location` anyway. If you wanna listen for changes though, there's the `emitter.on('pushState')` event that has the new location. I think that might perhaps be what you're looking for (:

With `nanocomponent` re-rendering based on new arguments can be expressed as:
```js
var nanocomponent = require('nanocomponent')
var html = require('html')

var myComponent = nanocomponent({
  render: function (state) {
    return html`<div>hi</div>`
  },
  onupdate: function (el, location) {
    console.log('new location was passed, mutate the DOM node!')
  }
})

module.exports = myView

function myView (state, emit) {
  return html`
    <section>
      ${myComponent(window.location)}
    </section>
  `
}
```

---
@bcomnes I'm thinking of including `nanocomponent` in the handbook, and try and keep the README minimal - only explain sibling reordering (tbi) and node caching (without using modules)

---

@timwis fair point about having an example dir; I'm mildly on the fence about it because I was trying to mimick the [hypercore](https://github.com/mafintosh/hypercore) directory layout... but yeah maybs we should just move it to a dir

---

example.js
yoshuawuyts: :+1:

example.js
yoshuawuyts: It's still there, and should be in the README!  - the reason for `app.mount()` is that I like writing my apps as:
```html
<body>
  <header></header>
  <main></main>
  <footer></footer>
</body>
```
`app.mount()` is to replace a DOM node. (e.g. `document.body`), `app.start()` is to append to a DOM node.

example.js
yoshuawuyts: Oh yeah this is for the tests to run

example.js
yoshuawuyts: Derp, yeah I should publish it, it's still just on my machine - 10 lines tho but yeah lol

example.js
yoshuawuyts: Fair point

example.js
yoshuawuyts: :+1: fair point

example.js
yoshuawuyts: Kinda copied this pattern from choo 4's `subscription` API - the subs were only ever booted after the app was loaded. I guess this also prevents weird things from starting if you ever run the code in Node

bcomnes:
The handbook is good too.

index.js
```diff
@@ -1,216 +1,118 @@
...
-    if (opts.href !== false) {
-      subs.handleHref = function (send, done) {
-        onHref(function navigate (location) {
-          send('location:set', location, done)
-        })
-      }
-    }
+  function toString (location, state) {
+    state = state || {}
+    return router(location, state)
```
finnp: This should return `router(location, state).toString()`, right?

index.js
```diff
@@ -1,216 +1,118 @@
...
-  function render (state, data, prev, name, createSend) {
-    if (!_frame) {
-      _frame = nanoraf(function (state, prev) {
-        const newTree = _router(state.location.href, state, prev)
-        _rootNode = yo.update(_rootNode, newTree)
-      })
-    }
-    _frame(state, prev)
+  function route (route, handler) {
+    router.on(route, function (params) {
```
finnp: Should this be `function (params, state, emit)` ? If so `nanorouter` would need to be patched: https://github.com/yoshuawuyts/nanorouter/pull/1

example.js
timwis: ooooh, right - funny, i did the same thing in an app i'm writing so that tests could run last week. maybe a comment saying "so tests can run"? lol, unless you think most folks wouldn't wonder about that.

example.js
timwis: oh, interesting. I didn't realise it was an either/or. I assumed you had to `mount` _and_ `start`. k thx

example.js
timwis: Interesting. I mean, most folks put their `<script>` tag at the bottom of `<body>` these days anyway, so it's already being run after most of the DOM is loaded, but I imagine if you're really trying to optimize, you'd want your JS to run even before your fonts are finished loading. (Though if you're _really_ trying to optimize, you'll probably use system fonts lol, but you get the point)

timwis:
At first glance, that `onupdate` stuff looks even easier to use than `prev`. @abradley2 what do you think? I don't have a chance to try it out today but hope someone will

index.js
yoshuawuyts: nah it's ok because we wrap it in the next line, no need to pass all those extra args

index.js
yoshuawuyts: ah lol, yeah :sparkles:

index.js
finnp: How does `toString` then passes the `state` argument around? Alternatively it could overwrite the state I guess.

yoshuawuyts: oh dang, you're right - yeah I was thinking the latter tbh. Given it's all sync it's safe

example.js
yoshuawuyts: good suggestion

yoshuawuyts:
Implemented all feedback!

bcomnes:
Are the recent changes released as a beta?

example.js
bcomnes: Its a cool pattern though.  Maybe we can note it in the handbook.

README.md
```diff
@@ -193,552 +158,87 @@ is laid out. And once an application is built, we want it to be small,
...
-* `template strings`
-
-This does not include Safari 9 or any version of IE. If support for these
-platforms is required you will have to provide some sort of transform that
-makes this functionality available in older browsers.  The test suite uses
-[es2020](https://github.com/yoshuawuyts/es2020) as a global transform, but
-anything else which might satisfy this requirement is fair game.
-
-Generally for production builds you'll want to run:
+`choo` uses [nanomorph][nanomorph], which diffs real DOM nodes instead of
```
bcomnes: fix link

bcomnes:
I found a few issues around input elements.  I opened issues at nanomnorph, but I'm thinking maybe they should live here.

https://github.com/yoshuawuyts/nanomorph/issues/50:

If you type into an input type=search element, that text is cleared on every render event.

https://github.com/yoshuawuyts/nanomorph/issues/48: 

If you try to bind an input=range element to state, it doesn't reflect state till you click on it, resize it, or directly set the value property on the dom node.

I've not been able to find the source of the bug on either of these.  Any ideas?

yoshuawuyts:
@bcomnes yeah, to get the latest pre-release run:
```sh
$ npm i choo@next
```

yoshuawuyts:
Ok merging to master but not releasing yet; we need to finish the checklist in #429 before we can release. I'll post here once we release (which should be this week I think?) Anyway, whooooooop here it goes! :sparkles:

yoshuawuyts:
:sparkles: :sparkles: :sparkles: thanks for all the feedback everyone! :sparkles: :sparkles: :sparkles:

bcomnes:
That was fun!

Issue #429, opened 2017-03-19
choo v5 release checklist


yoshuawuyts:
Choo `v5` is super close; this the stuff we need to do:

- [x] implement choo v5 API (#425)
- [x] update website
- [x] update handbook
- [x] probably just delete the rest of the handbook
- [x] update `choo-log` (branch ready)
- [x] update `choo-persist` (branch ready)
- [x] double check readme
- [ ] write patch notes
- [x] close like all issues

---

cc/ @bcomnes @timwis @toddself

bcomnes:
I'll try to dip my toes in when I finish up a few other things.

yoshuawuyts:
ok, added a warning to the handbook that it's out of date. Updated website, readme all the other stuff. Someone should write the patch notes (volunteers welcome!) but yeah hitting the release button right now

yoshuawuyts:
ps we got mentioned in the [node newsletter](https://twitter.com/yoshuawuyts/status/844192889341132802) and oops they said v5 is out yet - so yeah given v5 is pretty much done we can def release. Here goeessss

fyrkant:
Is there a upgrade guide from v4 -> v5 somewhere? Maybe someone updated a v4 app, leaving some nice juicy commit changes I could look through to get a hang of how things translate?

yoshuawuyts:
@fyrkant uhhh, we should probably write patch notes but yeah like almost everything changed haha - I wrote down some of the changes [in an article](https://medium.com/@yoshuawuyts/choo-v5-bc775b007b5e#.yqnw19ap7) tho, I hope it makes sense

tgfjt:
@yoshuawuyts I read the awesome article! but I felt "oh? where should I write `subscriptions[]`!?"
We would be happy with an upgrade guide or something example.:)

tunnckoCore:
@tgfjt that's the tricky part, choo v5 seems to not be ELM-like anymore, everything is just events.

After looking it few days, i actually don't like v5.

More hints: 

- models are defined through `.use` method, "plugins" are passed with `state, emitter` signature, where `state` is empty object and `emitter` is real event emitter 
- the old `send` from the views is now `emitter.emit` function, which should be used to emit "actions" defined with `emitter.on` method from inside the "plugin" (which now is model)

tunnckoCore:
@yoshuawuyts add "add tests" bullet to the list ;d

tgfjt:
[...]

PR #432, opened 2017-03-21
more API docs for choo v5

yoshuawuyts:
Document the router for v5. Ref #429

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@
       Website
     </a>
     <span> | </span>
-    <a href="https://yoshuawuyts.gitbooks.io/choo/content">
+    <a href="https://github.com/yoshuawuyts/choo-handbook">
       Handbook
     </a>
     <span> | </span>
--- a/README.md
+++ b/README.md
@@ -257,10 +257,28 @@ Choo fires messages when certain events happen:
 - __`.on('pushState')`__: when the history API is triggered
 
 ### `app.route(routeName, handler)`
-Register a route on the router
+Register a route on the router. Uses [nanorouter][nanorouter] under the hood.
+Params can be registered by prepending the routename with `:routename`, e.g.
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
```txt
16:36 <josephluck> are query parameters covered?
16:37 <runeh> query args can just be extracted with the usual window.location incantations can't they? As long as it's documented that that is the way to do it?
16:38 <runeh> anyway, those docs seem reasonable. Might mention explicitly that there will be a render on navigation? Since everywhere else you have to emit and explicit render event
16:38 <yoshuawuyts> runeh: good point
16:39 <yoshuawuyts> runeh: we do some custom logic to extract the params
16:39 <yoshuawuyts> don't think it matters too much how we do it tbh - people can read the source if they're interested; it's just a few hundred lines anyway
```

README.md
```diff
@@ -257,18 +257,27 @@ Choo fires messages when certain events happen:
...
+Register a route on the router. Uses [nanorouter][nanorouter] under the hood.
+Params can be registered by prepending the routename with `:routename`, e.g.
+`/foo/:bar/:baz`. The value of the param will be saved on `state.params` (e.g.
+`state.params.bar`). Wildcard routes can be registered with `*`, e.g. `/foo/*`.
+The value of the wildcard will be saved under `state.params.wildcard`.
+
+Using hashes to delimit routes is supported out of the box (e.g. `/foo#bar`).
+When a hash is found we also check if there's an available anchor on the same
+page, and will scroll the screen to the position. Using both hashes in URLs and
+anchor links on the page is generally not recommended.
```
lrlna: may or may not be good to mention working with query parameters too? idk if folks normally use those (i do, so mentioning it)

yoshuawuyts: The first paragraph mentions query params, think that should be enough?

PR #433, opened 2017-03-21
Fixes for example code

byronhulcher:
Hello! Just some small fixes that I think would improve the example code.

1) Missing dependencies
2) A bug preventing the proper todoItem from being deleted from the active and completed lists

Let me know if there's anything I can do to improve this PR!

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -131,11 +131,21 @@ function todoStore (state, emitter) {
 
     if (todo.done) {
       var done = state.todos.done
-      var doneIndex = done[todo]
+      var doneIndex
+      done.forEach(function (_todo, j) {
+        if (_todo.id === id) {
+          doneIndex = j
+        }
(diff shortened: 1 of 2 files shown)


byronhulcher:
Also while fixing this I noticed a small dev issue that I did not address in this PR as I'm not sure how you'd like to resolve it. Since you have `var choo = require('../')` which pulls in some dependencies (for exampe, nanomount) when you run `npm run deps` from within the example directory, it actually complains that the example/package.json is missing those deps

```
> choo-todomvc-example@1.0.0 deps /Users/bhulcher2/scratch/choo/example
> dependency-check . && dependency-check . --extra --no-dev

Fail! Dependencies not listed in package.json: document-ready, nanorouter, nanomount, nanomorph, nanoraf, nanobus, global
```

PR #436, opened 2017-03-23
remove the global dep

yoshuawuyts:
None of these files will ever run in Node, so there's no point in having them depend on `global`. Shave those bytesss :sparkles:

### DIFF
--- a/lib/history.js
+++ b/lib/history.js
@@ -1,5 +1,3 @@
-var document = require('global/document')
-var window = require('global/window')
 var assert = require('assert')
 
 module.exports = history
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
Did some quick benching and with `global` removed choo sits at a comfortable `4.03kb` 
screen shot 2017-03-23 at 10 16 33

yoshuawuyts:
screen shot 2017-03-23 at 10 38 48


`4.0.1kb` lol

sotojuan:
Awesome!

yoshuawuyts:
`v5.0.2`

PR #439, opened 2017-03-23
hash fix: wait to run scrollIntoView

sethvincent:
I'm working on porting [minidocs to choo v5](https://github.com/freeman-lab/minidocs/tree/choo-v5) and found a couple issues with hash behavior.

I'm open to revising things if needed, or even just using this PR as inspiration for better ones if there are other ways to solve things.

Here's a description of changes:

### `scrollIntoView`

I found that the check for an id was failing because it needed to wait for the next tick: https://github.com/yoshuawuyts/choo/blob/master/index.js#L113

This PR puts `scrollIntoView` inside a `setTimeout`.

> edit: removed `createLocation` changes

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,5 +118,7 @@ function scrollIntoView () {
 }
 
 function createLocation () {
-  return window.location.pathname.replace(/\/$/, '')
+  var pathname = window.location.pathname.replace(/\/$/, '')
+  var hash = window.location.hash.replace(/^#/, '/')
+  return pathname + hash
 }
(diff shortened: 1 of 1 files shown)


jekrb:
Just as an example, it looks like Next.js is wrapping it's `scrollIntoView` in a `setTimeout` here (commenting that it won't scroll properly otherwise) 

https://github.com/zeit/next.js/blob/97c472c0c8419d7d032c5f98e21123968b63b869/lib/app.js#L45-L53

sethvincent:
@jekrb ah, good to know other folks have come up against this.

I've revised this PR to be _just_ for wrapping `scrollIntoView` in a `setTimeout`, and will open another with a revised approach for handling hash routing.

yoshuawuyts:
`v5.0.3`

PR #442, opened 2017-03-24
Use npm-installed requires in README example

brechtpm:
To make it more copy-pastable for beginning users... ;)


yoshuawuyts:
Oh yeah good one, thanks!

PR #458, opened 2017-03-31
Fix crash in node

tornqvist:
Directly accessing properties on window crashes in node

Steps for reproduction

```
$ node
> require('./index')()
```


yoshuawuyts:
`v5.1.4`

PR #460, opened 2017-04-01
remove deopts from example

yoshuawuyts:
We were causing sync UI recalc by running:
```js
e.target.value = ''
```
Instead we now push the mutation to the start of the frame which makes things run smoothly. (now all I gotta do is figure out why input event cause deopts, but yeah that's it yayay)

screen shot 2017-04-01 at 14 02 03


yoshuawuyts:
Merging

PR #462, opened 2017-04-01
optimize example further

yoshuawuyts:
Expands on #460, debounces input & removes long frames from key events by pushing them to the start of the next frame. The remaining long frames are because of choo's diffing function in the case of large numbers of todos - which is something that can be optimized through a combination of Nanocomponent and https://github.com/yoshuawuyts/nanomorph/issues/8.

### screenshot

screen shot 2017-04-01 at 19 59 06


yoshuawuyts:
This should be good now, merging!

PR #465, opened 2017-04-04
call DOMContentLoaded on next frame

yoshuawuyts:
This delays emitting `DOMContentLoaded` until the start of the next frame. Should help squash long frames when booting. Apparently calling `window.requestAnimationFrame` as the direct result of the event pushes stuff to execute on the same frame. So hence wrapping it to run on the next tick.

### Before
There's a 68ms long frame because `DOMContentLoaded` fires and triggers more stuff in the same frame.

screen shot 2017-04-04 at 19 46 36

### After
It's now cleanly split into two smaller frames

screen shot 2017-04-04 at 19 46 05

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,7 +51,15 @@ function Choo (opts) {
   }
 
   function start () {
+    if (hasPerformance && timingEnabled) {
+      window.performance.mark('choo:renderStart')
+    }
     tree = router(createLocation())
+    if (hasPerformance && timingEnabled) {
+      window.performance.mark('choo:renderEnd')
+      window.performance.measure('choo:render', 'choo:renderStart', 'choo:renderEnd')
+    }
+
     rerender = nanoraf(function () {
       if (hasPerformance && timingEnabled) {
         window.performance.mark('choo:renderStart')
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
This feels like a hack tho - shouldn't merge until we test more

juliangruber:
I'm also asking myself if splitting this into two frames actually helps with user perceived performance

yoshuawuyts:
@juliangruber yes and no - this prevents a 60ms lock on the UI which is bad if you're doing animations (e.g. it'll visibly lag). On the other hand it does cause a delay in boot up performance because we're calling `setTimeout(fn, 0)` and `window.requestAnimationFrame` multiple times in a row. So it'll take longer for things to resolve. I guess it depends on what you're doing, but I think that not having any long frames is the best choice in most cases.

As substack pointed out in IRC we should probably consider hooking into the browser's microtask queue as it provides us with a Node timer-like API for frames - which is exactly what we want. [medikoo/next-tick](https://github.com/medikoo/next-tick) is probably the place to start.

yoshuawuyts:
Ok, found a way to make this even more efficient by hooking into the browser's microtask queue. Trace now looks clean, nice and done without any hacks. Will push in a bit.

screen shot 2017-04-05 at 00 29 50

yoshuawuyts:
Now uses https://github.com/yoshuawuyts/nanotask/tree/master. Makes it so when calling `.mount()` it's forced to run in the next frame which makes it so the initial frame only ever does the boot stuff we can't avoid. E.g. parse HTML, parse CSS, parse JS and kick off all corresponding processes.

Also found out that microtasks run at the end of the current tick (like `process.setTimeout`) and `setTimeout(cb, 0)` runs at the start of the next tick (like `setImmediate(cb)`). We're using that to our advantage here to make sure some stuff always resolves async, but as fast as possible - and rendering resolves at the start of the next frame. This should make for a nice spread of load, keeping the frame rate up.

yoshuawuyts:
Closed in favor of #489

PR #469, opened 2017-04-05
Provide an option for appending

toddself:
Modify `.mount` to accept an options hash to let users append the tree to the
selector rather than replace the selector


yoshuawuyts:
Hmmm, mixed feelings. Apart from people being confused (renaming the method might help with that) do we have a use case for this? When wouldn't people want to replace a Node? I think that'd mostly be in cases where choo is not an optimal fit; e.g. used in `<iframe>` or similar, where there's no need for routing and the like. Ghmmmm

yoshuawuyts:
Talked about this on IRC yest, closing

PR #473, opened 2017-04-11
pluggable render methods

jongacnik:
This adds the ability to swap out nanomorph/nanomount with your own renderer via `app.use()`. The following is an example of using [preact](https://github.com/developit/preact) to render your choo app:

```js
var choo = require('choo')
var preact = require('preact')
var hyperx = require('hyperx')
var html = hyperx(preact.h)

var app = choo()
app.use(preactRenderer)
app.route('/', mainView)
app.mount('body')

function preactRenderer (state, emit, renderer) {
  renderer.render = function (tree, newTree, root) {
    return preact.render(newTree, root, tree)
  }
  renderer.mount = function (tree, root) {
    return preact.render(tree, root)
  }
  renderer.toString = function (html) {
    return someRenderToStringMethod(html)
  }
}

function mainView (state, emit) {
  return html`
    <body>
      <h1>Helloooo from preact!</h1>
    </body>
  `
}
```

`renderer` is passed as an additional, third parameter to `app.use()` where you can define your own `render`, `mount`, and `toString` methods. 

Perhaps this kind of addition falls outside the scope of choo's broader goals, but I found it suuuper cool recently when I had an app written w/ choo, I ran into needing to use vdom towards the end of the build, and was able to swap out the renderer while making 0 changes to the rest of the code! ✨

Best, J

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,7 +26,13 @@ function Choo (opts) {
   var bus = nanobus()
   var rerender = null
   var tree = null
+  var root = null
   var state = {}
+  var renderer = {
+    render: handleRender,
+    mount: handleMount,
+    toString: handleToString
+  }
 
   return {
     toString: toString,
@@ -46,7 +52,7 @@ function Choo (opts) {
   }
 
   function register (cb) {
-    cb(state, bus)
+    cb(state, bus, renderer)
   }
 
   function start () {
@@ -56,7 +62,7 @@ function Choo (opts) {
         window.performance.mark('choo:renderStart')
       }
       var newTree = router(createLocation())
-      tree = nanomorph(tree, newTree)
+      tree = renderer.render(tree, newTree, root)
       if (hasPerformance && timingEnabled) {
         window.performance.mark('choo:renderEnd')
         window.performance.measure('choo:render', 'choo:renderStart', 'choo:renderEnd')
@@ -102,10 +108,9 @@ function Choo (opts) {
   function mount (selector) {
     var newTree = start()
     documentReady(function () {
-      var root = document.querySelector(selector)
+      root = document.querySelector(selector)
       assert.ok(root, 'could not query selector: ' + selector)
-      nanomount(root, newTree)
-      tree = root
+      tree = renderer.mount(newTree, root)
     })
   }
 
@@ -113,8 +118,21 @@ function Choo (opts) {
     state = _state || {}
     var html = router(location)
     assert.equal()
-    return html.toString()
+    return renderer.toString(html)
   }
+}
+
+function handleRender (tree, newTree, root) {
+  return nanomorph(tree, newTree)
+}
+
+function handleMount (tree, root) {
+  nanomount(root, tree)
+  return root
+}
+
+function handleToString (html) {
+  return html.toString()
 }
 
 function scrollIntoView () {


jongacnik:
fwiw here are a couple examples of renderers 🤘

[virtual-dom](https://github.com/Matt-Esch/virtual-dom)

```js
var vdom = require('virtual-dom')
var hyperx = require('hyperx')
var html = hyperx(vdom.h)
var rootNode

function vdomRenderer (state, emit, renderer) {
  renderer.render = function (tree, newTree, root) {
    var patches = vdom.diff(tree, newTree)
    rootNode = vdom.patch(rootNode, patches)
    return newTree
  }
  renderer.mount = function (tree, root) {
    rootNode = vdom.create(tree)
    root.appendChild(rootNode)
    return tree
  }
}
```

[Preact](https://github.com/developit/preact)

```js
var preact = require('preact')
var hyperx = require('hyperx')
var html = hyperx(preact.h)

function preactRenderer (state, emit, renderer) {
  renderer.render = function (tree, newTree, root) {
    return preact.render(newTree, root, tree)
  }
  renderer.mount = function (tree, root) {
    return preact.render(tree, root)
  }
}
```

[React](https://github.com/facebook/react)

```js
var React = require('react')
var ReactDOM = require('react-dom')
var hyperx = require('hyperx')
var html = hyperx(React.createElement)

function reactRenderer (state, emit, renderer) {
  renderer.render = function (tree, newTree, root) {
    return ReactDOM.render(newTree, root)
  }
  renderer.mount = function (tree, root) {
    return ReactDOM.render(tree, root)
  }
}
```

[Snabby](https://github.com/jamen/snabby)

```js
var html = require('snabby')

function snabbyRenderer (state, emit, renderer) {
  renderer.render = function (tree, newTree, root) {
    html.update(tree, newTree)
    return newTree
  }
  renderer.mount = function (tree, root) {
    html.update(root, tree)
    return tree
  }
}
```

[Inferno](https://github.com/infernojs/inferno/)

```js
var Inferno = require('inferno')
var hyperx = require('hyperx')
var html = hyperx(require('inferno-create-element'))

function infernoRenderer (state, emit, renderer) {
   renderer.render = function (tree, newTree, root) {
    return Inferno.render(newTree, root)
  }
  renderer.mount = function (tree, root) {
    return Inferno.render(tree, root)
  }
}
```

yoshuawuyts:
Heya 👋, thanks for the PR! Cool to see more of the Folder studio folks drop by :grin: I've taken some time to think about and talk to other people about it. We think this PR might not be ideal for `choo` for the following reasons:

### Configuration
If we allow configuring the renderer, for consistency it would make sense to allow configuration for the diffing engine, event bus and router too. We think that might be a bit much; there'd be so little left in the framework.

Also by making things pluggable rather than building them from scratch you include a whole bunch of unused code. Dead code elimination could help with this, but that means people would need to start jumping through hoops because of a problem we created - where we could just not create the problem.

### Maintenance
We'd need to ensure choo stays compatible with all the libs you mentioned, which means maintainers should have knowledge of those libs and be able to triage interop problems. I really don't feel like doing that haha.

### Ecosystem compatiblity
It becomes tricky to brand peer dependencies as "built for choo" if "choo" can mean any number of things. They'd need to change to "built for choo with the default renderer" or the like. This is trickier to communicate than "works with choo".

To counter this argument: we've made API changes in the past, so not everything might be compatible still - but we're committed to not break the API anymore so this will most likely not be a problem going forward. It'd be cool if the "choo" brand carried certain meaning and could set people's expectations.

### Ease of forking
Choo itself is about 100 lines of glue; something that can be written in less than an afternoon. I think if people want to see a {react,preact,inferno} version of choo, they could easily roll their own. In essence all we're doing is binding an event emitter to a render engine and router, this is closer to a paradigm than an exact library. It'd be cool to see different flavors of this paradigm using different libraries and engines :sparkles:

---

Hope this all makes sense. We really appreciate the time and effort you put into this patch tho, sending in a patch is always a great way to start a discussion! Cheers -Yosh

josephluck:
@jongacnik: Sorry to dive in here, but I've been working on something that achieves custom rendering: https://github.com/josephluck/helix might be of relevance?

jongacnik:
@yoshuawuyts: Of course! These are all great points. I think I had a little bit of: "oh, this can mayyyybe just be an invisible feature, we don't need to document, shhhh, there if you need it 👻" – circumventing the need to _really_ worry about things like compatibility/dce. But, 100% understand keeping that can of worms closed. Also hadn't put it specifically in the lens of "built for/with choo" which makes a lot of sense. The commitment to morphin' dom nodes is important, I think.

Either way, it's a good exercise and happy to have the convo. Maybe this thread can serve as a little reference for people who might want, as you said, to roll their own iterations of the paradigm. Aaand, for anyone interested, [rooch](https://github.com/yoshuawuyts/rooch) (`preact` + `choo`) was recently brought up to choo v5 style.

Thanks bunches for the consideration and thorough response ~ choo onward!

PR #476, opened 2017-04-14
use nanotiming

yoshuawuyts:
- use `nanotiming` which allows reducing the bundle size at the application level by reusing all the timing stuff (https://github.com/yoshuawuyts/nanotiming)
- breaking change
- so don't merge right now


yoshuawuyts:
Also worth noting is that we should probably move the clear timing buffer logic from `choo-log` into `choo`

yoshuawuyts:
Closing in favor of #489

PR #479, opened 2017-04-18
run choo event handlers first

### LINKED ISSUE #463: Params not yet available in 'pushState' and 'render' event handlers
### Expected behavior

When handling a 'pushState' event, the resolved `state.params` should already be available upon calling the event handler.

### Actual behavior

Resolved params are only added to state during render.

To reproduce, click on the link to 'some page' in this example app:

```js
var choo = require('choo')
var html = require('choo/html')

var app = choo()

app.route('/', function () {
  return html`
    <body>
      <a href="/somepage">some page</a>
    </body>
  `
})

app.route('/*', function (state) {
  return html`
    <body>
      page: ${state.params.wildcard} <!-- correctly renders 'somepage' -->
    </body>
  `
})

app.use(function (state, bus) {
  bus.on('pushState', function () {
    console.log(state.params.wildcard) // logs undefined
  })

  bus.on('render', function () {
    console.log(state.params.wildcard) // logs undefined
  })
})

app.mount('body')
```

This is a problem for my specific use case, because in the 'pushState' (or 'render') event handler, I want to fetch some data based on the page param, and fill out the page based on the response.

As an aside, I do want to mention that working with `choo v5` has been a real joy so far. You absolutely nailed it with the new stripped-down API for `app.use`!

yoshuawuyts:
Fixes #463 by letting choo handle events before any other handlers run (unless other handlers explicitly want to run before choo through `.prependListener()` instead of `.on()`). Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,14 +62,14 @@ function Choo (opts) {
       }
     })
 
-    bus.on('render', rerender)
+    bus.prependListener('render', rerender)
 
     if (opts.history !== false) {
       nanohistory(function (href) {
         bus.emit('pushState')
       })
 
-      bus.on('pushState', function (href) {
+      bus.prependListener('pushState', function (href) {
         if (href) window.history.pushState({}, null, href)
         bus.emit('render')
         setTimeout(function () {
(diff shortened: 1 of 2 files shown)


toddself:
lgtm

yoshuawuyts:
`v5.2.0` :tada:

PR #480, opened 2017-04-20
become prototype based

yoshuawuyts:
Was wondering if we could make Choo prototype based and turns out we totally can. Think I like the internals a bit better this way. In addition makes `choo.state`, `choo.emitter` and `choo.router` public things people can access in case they wanna do stuff with it (not necessarily encourage, but yeah no reason to hide it either). This is definitely a breaking change, although the API doesn't change much.

Size went up slightly to `4.07KB`.

Keen to hear thoughts on this!

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ var assert = require('assert')
 module.exports = Choo
 
 function Choo (opts) {
+  if (!(this instanceof Choo)) return new Choo(opts)
   opts = opts || {}
 
   var routerOpts = {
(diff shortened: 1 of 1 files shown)


tunnckoCore:
it's a lot better and clean now, yea

tornqvist:
Though exposing some of the internals (e.g. `state`, `emitter`, and `router`) does make sense, I'm personally not a big fan of the prototype pattern unless you're going for extensibility, which I don't really see the benefit of in the case of choo. Having to deal with bindings of `this` and all those underscored properties hanging around makes for more clutter, imo.

marlun:
I agree with @tornqvist I think using prototypes when not necessary only adds clutter (complexity). It's not like prototypal inheritance is known for being easy to understand. Before it was just functions and scope and at least for me that feels easier and cleaner to understand for someone interested about the internals. Might be a good reason to ignore this and it might be that I'm wrong and that people think that prototypes are easier to understand :)

tunnckoCore:
Prototypes, prototypes, prototypes... Stop this.. i don't know what.
It's not that hard to understand what happens, hence see the diff, such small change is nothing. Everyone know `this` and everyone know what prototype is. And no one need to know that Choo is old school class.

Really, it can be done with more elegance and without `this` and prototypes, but that's how Yosh and the party around him like to write code - everywhere some references and such things, and because of that it make sense to be using prototypes directly.

yoshuawuyts:
> Everyone know this and everyone know what prototype is. And no one need to know that Choo is old school class.

@tunnckoCore Though you're right that people generally don't need to bother with internals, I'm not convinced everyone understands how prototypes work.

---

>  I'm personally not a big fan of the prototype pattern unless you're going for extensibility, which I don't really see the benefit of in the case of choo.

@tornqvist a fun use case opens up by doing this: people get to define their own router. Which is not interesting if you're building regular apps - but if, say, you're targeting iframes it can dramatically simplify your routing code. With JS being multi-paradigm prototypes aren't a requirement, but I feel it makes for the best fit.

Also another thing that's interesting is that people can create their own initial state, which is useful if you're doing IPC stuff with Electron. They have this shared object thing that just works between processes, but for it to integrate with choo it would need to be the object we use for `state` - simply copying values would not cut it.

bengourley:
_Slow to get around to this but saw it in my inbox and wanted to weigh in as I'm a pretty heavy user of choo!_

+100 for moving to prototypal pattern. Being able to poke around with the prototype makes it easier to monkey patch stuff, and makes the following use cases much better: 

1. being able to tweak/hack on changes without forking (the hidden internals made it really hard to hack on the [hot reload](/bengourley/choo-resume) stuff)
1. making necessary, context-specific tweaks/changes in an existing choo app (including pre-patching a bug that might be fixed in a later version)

yoshuawuyts:
Closing this in favor of https://github.com/yoshuawuyts/choo/pull/489 (merged into a separate branch) :D

PR #482, opened 2017-04-24
expose router

yoshuawuyts:
In preparation for GH-480, but also because choo 4 exposed the router, and we need it for server rendering


yoshuawuyts:
`v5.3.0` :tada:

Issue #484, opened 2017-04-25
Choo v5 fails to replace route HTML


mantoni:
Thanks for Choo 5! It's a great update. I like it ❤️ 

Apparently, I can no longer have a route with something like this:

```js
if (state.condition) {
  return html`This`
}
return html`That`
```

This used to work in Choo v4. I'm currently migrating to v5 where this does nothing when the condition changes.

I've set up a repository to demo the issue: <https://github.com/mantoni/choo-replace-page-html-issue>
Setup instructions are in the readme.

yoshuawuyts:
Ah yeah, this is because the DOM nodes are being unmounted when you re-render, we should have better warnings for this. So DOM node morphing happens like this:
- compare two nodes
  - if the nodes are the same type, copy all attributes from one to the other
  - if nodes are different types, create a new node

Because the root node was being recreated, it's suddenly unmounted from the DOM, and it appears as if nothing changed. Oops :cry:

yoshuawuyts:
Documentation fix in #485, closing this issue. Thanks, and apologies you ran into this.

mantoni:
This should either be fixed or throw an exception then. It's very confusing, because my example works if the two code branches happen to return the same type of nodes. No?

mantoni:
Ah, sorry. The runtime assertion would throw in my case. That's fine then.

PR #485, opened 2017-04-26
warn on unmount

yoshuawuyts:
Adds a runtime assertion to warn if the root node is being unmounted. Should help catch cases like https://github.com/yoshuawuyts/choo/issues/484. Perhaps this should be caught at the `nanomorph` layer, although: perhaps not. Anyway, hope this helps. Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,6 +57,7 @@ function Choo (opts) {
       }
       var newTree = router(createLocation())
       tree = nanomorph(tree, newTree)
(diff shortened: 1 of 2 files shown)


mantoni:
👍 I can confirm this throws in case of the example provided in #484. Thanks.

yoshuawuyts:
`v5.4.0` :tada:

PR #493, opened 2017-05-19
Allow listeners to attach before rendering

emilbayes:
This will allow `choo-redirect` to emit `pushState` on the first render. I'm not sure how to test this though.


emilbayes:
So the important points here are:

* Attach pushState listeners before render
* Attach the `render -> rerender` listener before firing the first `render` through `route`

Issue #470, opened 2017-04-06
emitter/router doesn't work immediately


blahah:
in v5, if I have a function like so...

```js
app.use((state, emitter) => {
  emitter.emit('pushState', '/someroute')
})
```

it doesn't do anything.

Whereas if I do:

```js
app.use((state, emitter) => {
  setTimeout(emitter.emit('pushState', '/someroute'), 200)
})
```

it works.

It feels as though the `app.use()` callbacks shouldn't be called until the api is ready to be used.

yoshuawuyts:
ya you should probably wait for `emitter.on('DOMContentLoaded')` before
firing these events like as a general rule - that way they always fire
after start() has taken off

blahah:
@yoshuawuyts is there any reason why the `.use()` functions should be loaded before then? If the app just waited for `DOMContentLoaded` before calling the callbacks, then the user doesn't have to worry about waiting for it explicitly.

If \o isn't going to happen, the need to wait for the event should definitely be documented.

yoshuawuyts:
.use() functions should be immediate so default data can be setup; views
would otherwise become littered with checks if properties exist. You're
right, we should document this better - PR would be very welcome here!

yoshuawuyts:
Closed in #472

Tyler-Murphy:
Could `pushState` and `replaceState` events be queued by choo until `documentReady`?

It's nice not to think about when the DOM is ready. Choo already hides this detail for us in [`mount`](https://github.com/yoshuawuyts/choo/blob/8ff7deea3cc0b1041338068211b3cf5662df8b2e/index.js#L109).

PR #504, opened 2017-05-28
Wraps updateHistory calls in documentReady

Tyler-Murphy:
This allows 'setState' and 'replaceState' to be used in models without checking whether the document is ready.

Since `documentReady` runs callbacks on the next tick if the document is already ready, this does change behavior slightly. It seems like it's fine to wait a tick, but I haven't investigated thoroughly.

Related to #470, which was solved by documenting the fact that history changes don't work before the document is ready.


Tyler-Murphy:
Whoops, it actually looks like https://github.com/yoshuawuyts/choo/pull/493 added this behavior. But this part of the documentation was missed?

> Routing via pushState will not work until the DOMContentLoaded event has been fired.

yoshuawuyts:
Docs PR welcome if this was confusing (:

PR #508, opened 2017-05-31
Allow router to accept an object of routes.

smcmurray:
Define multiple routes with a single method call:

    app.route({
      '/': main
      ,'/other': (state, emit)=>html`some other content`
    })

Also, allow chaining:

    app.route('/', main)
          .route('/other', (state, emit)=>html`some other content`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,16 +37,23 @@ function Choo (opts) {
   }
 
   function route (route, handler) {
-    router.on(route, function (params) {
-      return function () {
-        state.params = params
-        return handler(state, emit)
-      }
-    })
+    if ('object' === typeof route){
+      Object.keys(route).forEach(k=>route(k, route[k])
+    }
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Hey, I've thought about this PR a while, and ultimately don't think this is the right thing to add to core as it's easy enough to implement in userland. Thanks for your PR tho; I hope choo is working out well for you so far! C:

smcmurray:
@yoshuawuyts Your response seems to me to be appropriate for the object of routes idea. I ask you, though, to reconsider the usefulness of chaining choo methods, including the route method. Such a change would not break any current functionality, either.

yoshuawuyts:
Yeah I like that! Tho probably not for .start() and .mount() because those
should be called last. PR welcome! 🎉✨

PR #489, opened 2017-05-08
v6 beta branch

yoshuawuyts:
in-progress target branch we can use for breaking behavior. The prototype branch is basically a full rewrite; this allows us to put out some betas and iterate on the minor changes we want for the next version.

Basically I'm thinking we keep the API stable, but yeah there's a few nits that are technically breaking. So we should iron those out in a semver major. Yay for boring changes :sparkles:

### related issues
- [x] call DOMcontentloaded on next frame https://github.com/yoshuawuyts/choo/pull/465
- [x] use nanotiming https://github.com/yoshuawuyts/choo/pull/476
- [x] become prototype based https://github.com/yoshuawuyts/choo/pull/480
- [x] add tracing to nanobus https://github.com/yoshuawuyts/nanobus/pull/10
- [x] revert special casing of root onload handlers https://github.com/yoshuawuyts/nanomorph/pull/65
- [x] fix select tags https://github.com/yoshuawuyts/nanomorph/pull/60
- [x] remove onload https://github.com/shama/bel/pull/69
- [x] should clear performance timers https://github.com/yoshuawuyts/choo/issues/461
- [x] remove nanomount https://github.com/yoshuawuyts/choo/issues/481
- [x]  remove nanohistory (too small a module)
- [x] fix whitespace in `bel`
- [ ] fix whitespace in `yo-yoify` https://github.com/shama/yo-yoify/issues/49
- [ ] fix whitespace in `pelo` https://github.com/shuhei/pelo/issues/7
- [ ] upgrading bankai to new `yo-yoify`
- [ ] upgrading `bel` to new `pelo`

PR #516, opened 2017-06-21
docs(README): remove list optimisation text

Charlotteis:
✨ ✨ ✨ 

Lists have been optimised via https://github.com/yoshuawuyts/nanomorph/issues/8 in `v5.0.0`
I've removed the text.

It may be that you still want to explain list reordering optimisations using nanomorph. If so, you can mash that close button and write up an explanation 📝 

Ty for choo! 🚂 

✨ ✨ ✨

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,9 +166,6 @@ Choo is reasonably fast out of the box. But sometimes you might hit a scenario
 where a particular part of the UI slows down the application, and you want to
 speed it up. Here are some optimizations that are possible.
 
-### Reordering lists
-To be implemented. (See [yoshuawuyts/nanomorph#8](https://github.com/yoshuawuyts/nanomorph/issues/8))
-
 ### Caching DOM elements
 Sometimes we want to tell the algorithm to not evaluate certain nodes (and its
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Haha, hey! Happy to see you here! :grin: - umm, so the thing is that `nanomorph` v5 is currently a lil broken (oops); work is being done in https://github.com/yoshuawuyts/nanomorph/pull/71 to fix it. That's why it's not part of any of the choo releases yet, and why that notice was still there haha.

That patch is one of two holding up the choo v6 release (see #489 for progress); the other patch is redoing https://github.com/shama/bel/pull/77 to fix whitespace issues in the `v4` line (also not shipped into choo yet).

Hope this makes sense haha :sparkles: - you're right in that we should write an explanation on how child reordering works tho; on the choo 6 branch I don't think we have written that yet. If you're keen a PR would be rad.

Anyway, yay thanks for the PR!

Charlotteis:
I would have no idea what to write! So I will not be able to 😂 

Thanks yosh~

PR #521, opened 2017-06-29
remove default route, parse qs

yoshuawuyts:
- removes the default route in favor of `'*'` (as per @timwis sugestion)
- documents the router
- add queryString parsing under `state.query`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ events: `'navigate'` will trigger whenever a route changes, `'replaceState'`
 can be called to redirect routes, and `popState` which is emitted when the back
 button in the browser is pressed.
 
+To top things off, we've reintroduced querystring parsing. An object containing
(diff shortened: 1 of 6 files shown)


timwis:
Woo! Should we mention that the wildcard should be the last route do it doesn't get triggered early?

yoshuawuyts:
> the wildcard should be the last route do it doesn't get triggered early?

Errr, does it? I believe you implemented it on the router, but with a trie structure the order of declaration shouldn't matter - if it does, it sounds like a bug on the router :/

yoshuawuyts:
Related to https://github.com/yoshuawuyts/choo/issues/519

yoshuawuyts:
Yeah, just verified in `wayfarer` that order indeed doesn't matter

yoshuawuyts:
Yeah it's b/c pelo needs to be updated - merging!

yoshuawuyts:
:package: `v6.0.0-1` - `npm i choo@next`

PR #529, opened 2017-07-09
State in tostring

yoshuawuyts:
Found a few bugs with server rendering.
- `state.title` wasn't being set on first render
- `state.events` wasn't being set when rendering on server.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ var nanohref = require('nanohref')
 var nanoraf = require('nanoraf')
 var nanobus = require('nanobus')
 var assert = require('assert')
+var xtend = require('xtend')
 
 module.exports = Choo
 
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
cc/ @choojs/trainspotters review welcome!

lrlna:
omg. yosh. that group name.

toddself:
`toString` is only used on the server end right?  Is that API available in the client? (Should it be used in the client)?  Should we tell browserify to not include xtend in the bundle?

Then again xtend is minuscule...

yoshuawuyts:
xtend is already being used by child deps, so it'll only use a single
instance. You're right tho - it shouldn't be used in the client b/c no
useful results will come out of it. Adding an assertion!

Issue #531, opened 2017-07-18
toString() not receiving state


louiscenter:
### Expected behavior
When running `app.toString('/', {animal: 'cat'})`, the `animal` property should become available in the application's `state` object.

### Actual behavior
The `animal` property is *not* available in the application's `state` object.

### Notes
`this.state` doesn't seem to be extended (with `xtend`) to include the `state` object passed in from `toString()`: 
https://github.com/choojs/choo/blob/a703ac5c56960193b0a06560f4bfa111e1b181e2/index.js#L179

louiscenter:
Should also mention i'm using version `6.0.0-3`

louiscenter:
This has now been resolved in `v6.0.0-4` ✨

PR #533, opened 2017-07-19
fix .toString() state (#531)

louiscenter:
Fixed a bug where `.toString()` method was not extending the `state` object properly: https://github.com/choojs/choo/issues/531

### DIFF
--- a/index.js
+++ b/index.js
@@ -176,7 +176,7 @@ Choo.prototype.mount = function mount (selector) {
 }
 
 Choo.prototype.toString = function (location, state) {
-  this.state = xtend(this.state || {}, state || {})
+  this.state = xtend(this.state, state || {})
 
   assert.notEqual(typeof window, 'object', 'choo.mount: window was found. .toString() must be called in Node, use .start() or .mount() if running in the browser')
   assert.equal(typeof location, 'string', 'choo.toString: location should be type string')
--- a/index.js
+++ b/index.js
@@ -176,7 +176,7 @@ Choo.prototype.mount = function mount (selector) {
 }
 
 Choo.prototype.toString = function (location, state) {
-  this.state = xtend({ events: xtend(this._events) }, this.state || {})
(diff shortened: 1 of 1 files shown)


louiscenter:
I have updated the fix from:

```js
this.state = xtend(state, this.state || {})
```

to

```js
this.state = xtend(this.state || {}, state || {})`
```

I found that if `state` was passed in as the first argument of `xtend()`, `this.state` wouldn't be updated properly if there were successive calls to `app.toString()` in the same node script.

My understanding is that the properties of `xtend()`'s second argument will take precedent over the first.

yoshuawuyts:
Oh yeah, now I remember - the approach here might leak values from previous
runs which is not good :/ - should find a way to create a new state on each
run - perhaps reset the state to the original state after the string has
been created or smth?

louiscenter:
would something like this work?

```js
Choo.prototype.toString = function (location, state) {
  this.state = xtend(this.state || {}, state || {})

  assert.notEqual(typeof window, 'object', 'choo.mount: window was found. .toString() must be called in Node, use .start() or .mount() if running in the browser')

  this.state.query = nanoquery(location)
  var html = this.router(location)
  assert.ok(html, 'choo.toString: no valid value returned for the route ' + location)

  var str = html.toString()
  this.state = { events: this._events }

  return str
}
```

index.js
```diff
@@ -176,7 +176,7 @@ Choo.prototype.mount = function mount (selector) {
 }
 
 Choo.prototype.toString = function (location, state) {
-  this.state = xtend({ events: xtend(this._events) }, this.state || {})
+  this.state = xtend(this.state || {}, state || {})
```
yoshuawuyts: `this.state` always exists; think we should be able to get away with `xtend(this.state, state || {})`

yoshuawuyts:
`v6.0.0-4`

PR #534, opened 2017-07-24
Save location

yoshuawuyts:
`window.location.href` is not available on servers, and when building components it's not uncommon to make decisions based on the current URL. Given we already compute the current href, this allows us to access it from our views. Thanks!


yoshuawuyts:
`v6.0.0-4`

PR #545, opened 2017-08-10
Add typescript definition

toddself:
Don't want to bikeshed on whether people like TypeScript or not :).

But here is a defintion file that TypeScript users can pick up and use.

I'm working on one for nanocomponent and nanobus (when the nanobus one is done
I'll update this PR so that it uses that instead of the node EventEmitter
definition, although I think they'll be the same...)

The question is do we want this in the project (adds a few K at best, and will
be ignored when the project is browserified), or should I add this to
DefinitelyTyped.

I totally prefer the typing to be in the file since there's no extra dep that
has be to be installed.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,38 @@
+/// <reference types="node" />
+
+import * as EventEmitter from 'events'
+
+export = Choo
+
+declare class Choo {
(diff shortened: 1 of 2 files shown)


graforlock:
*thud* *thud* are we merging it?

Issue #530, opened 2017-07-13
State out of date when switching routes


jondashkyle:
### Expected behavior
When attaching a callback to `state.events.NAVIGATE`,  and navigating to a new route, I expect `state.route` to represent the current route.

### Actual behavior
Instead, `state.route` is out of date. It should displaying the current route, but instead shows the previous. For example, if navigating from a route called `/cute` to one called `/nice`, it will return `/cute`.

```js
function model (state, emitter) {
  emitter.on(state.events.NAVIGATE, function () {
     console.log(state.route)
  })
}
```

yoshuawuyts:
Good catch! - think the fix for this would be to update `state.route`
whenever we call navigate

yoshuawuyts:
_ah!_ ok so this is tricky, because we can't know the route _until after_ we navigate. Why? Because it's the router that figures out which route we'll be using to render. The `'navigate'` event only triggers the router.

So the solution for this is to do:
```js
function model (state, emitter) {
  emitter.prependListener(state.events.NAVIGATE, function () {
    emitter.once(state.events.RENDER, function () {
       console.log(state.route)
    })
  })
}
```

The flow will then be:
```txt
navigate triggered -> navigate event fired -> model triggered ->
model render event attached (once) -> render event fired ->
router triggered -> new route set on state -> render complete ->
model render event triggered
```

Does this make sense? Hope it does haha; wish we could simplify it, but not sure how we could do this (besides, like, writing a module). Thanks!

akiva:
This is the same issue I am encountering in #538 , using `choo@next`, and the suggested fix doesn't actually work.

akiva:
OK, while the suggested fix didn't work for me, _this_ does:

```javascript
const app = require('choo')()
const html = require('choo/html')

app.use(function (state, emitter) {
  state.loggedIn = false

  emitter.on('DOMContentLoaded', function () {
    if (state.route !== '/shazam' && !state.loggedIn) {
      emitter.emit('pushState', '/shazam')
    }
  })

  emitter.prependListener('DOMContentLoaded', function () {
    if (state.route !== '/shazam' && !state.loggedIn) {
      emitter.emit('pushState', '/shazam')
      emitter.emit('render')
    }
  })

  emitter.prependListener('navigate', function () {
    console.log('state.route:', state.route, window.location.pathname)
    state.route = window.location.pathname
    emitter.once('render', function () {
      console.log('state.route:', state.route, window.location.pathname)
    })
  })

[...]

PR #553, opened 2017-08-30
Fire route event after render if the url has changed

jonjaques:
I believe this addresses #530 by adding an additional event name `route`. I think we sort of need a new event anyways to maintain compatibility.

This is perhaps a naive implementation, but just wanted to get this out there to get feedback.

I have tried moving the event emit around inside this function and I found the setTimeout to be necessary as I believe we need to wait for the internal router events to fire.

Anyways, let me know what you think!

P.S. Happy to add to docs too. I tried the test suite, but was unable to build out of the box on Windoze.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,8 @@ function Choo (opts) {
     PUSHSTATE: 'pushState',
     NAVIGATE: 'navigate',
     POPSTATE: 'popState',
-    RENDER: 'render'
+    RENDER: 'render',
+    ROUTE: 'route'
   }
 
   // properties for internal use only
@@ -132,11 +133,19 @@ Choo.prototype.start = function () {
 
   this.emitter.prependListener(self._events.RENDER, nanoraf(function () {
     var renderTiming = nanotiming('choo.render')
+    var oldHref = self.state.href + ''
 
     self.state.href = self._createLocation()
-    var newTree = self.router(self.state.href)
-    assert.ok(newTree, 'choo.render: no valid DOM node returned for location ' + self.state.href)
 
+    var newTree = self.router(self.state.href)
+
+    if (oldHref !== self.state.href) {
+      setTimeout(function () {
+        self.emitter.emit(self._events.ROUTE)
+      }, 1)
+    }
+
+    assert.ok(newTree, 'choo.render: no valid DOM node returned for location ' + self.state.href)
     assert.equal(self._tree.nodeName, newTree.nodeName, 'choo.render: The target node <' +
       self._tree.nodeName.toLowerCase() + '> is not the same type as the new node <' +
       newTree.nodeName.toLowerCase() + '>.')


yoshuawuyts:
@jonjaques hey! Thanks for putting together this PR. I was wondering if you could explain in a little more detail what the goal is here? How is this different from the `'navigate'` event? Thanks!

jonjaques:
Sure! Sorry I wasn't more specific initially 😅 

As outlined in #530, the `navigate` event fires off before application state is updated to the new route. Perhaps that is a bug, perhaps that is intended - but my goal with the `route` event is to be able to drive data loading without having to imperatively wire up stuff via `onload()` or something similar. IMHO if routing is handled by the framework, then we shouldn't have to add application code to patch up state that is owned by the framework. Just my 2 cents though 😄 

Another way to look at this would be that `navigate` signals before the page has rendered, and `route` signals after, however I'm open to any suggestions in that regard, because I would really like to avoid rendering an 'empty' page in between the time the new route renders and data coming back.

Here's a little proof of concept plugin that I've been working on: [choo-conductor](https://www.npmjs.com/package/choo-conductor)

jonjaques:
One thought on the above—perhaps we could expose an event object to be passed into the nav event? This would allow the user or a plugin to defer rendering in order to facilitate waiting for data or to provide some nice CSS transitions?

```
emitter.on('navigate', e => {
  if (state._transitioning) {
    state._transitioning = false
    return
  }
  e.preventDefault()
  state._transitioning = true
  doSomeTransitionMagic(x => emitter.emit('navigate', /* what to pass here??? */))
})

emitter.on('route', e => {
  cleanupTransition()
})
```

By the way, after an adjustment period (from React) - I'm really digging the framework. I like the ease of use for wiring up domain/app events and just the general idea (and power) of just rendering pure html and ditching all the bloat from virtual dom.

jonjaques:
@yoshuawuyts Just wondering if you have had a chance to review this behavior or have any thoughts regarding a resolution?

johnelliott:
I am also wondering if this is the problem with my app.

My app works properly when entirely server-rendered by turning on server rendering and breaking the script tag to my `bundle.js`. The links on the different pages just request from the server, the server builds the page from the rote and everything is happy without any client-side javascript.

My app works properly when clicking links around the app when client-side rendering. If I turn off server rendering of the app, I see a brief frame before the fetch of API data fills the app with content. The navigation works and only static images are requested to the server.

The problem I am having seems to be navigating in and out of the app to other sites or static files:

If I start at the index route `/`, then navigate via an `<a>` to another route of the app—`/view/image.jpg`—then click another `<a>` link, taking me to a static original file `/static/image.jpg` that's served from my static server I get unexpected behavior:

I expect that clicking the back button goes back to `/view/image.jpg`.

But actually clicking the back button goes to `/`.

I've really struggled understanding what the intended approach to building apps with Choo is because there aren't many examples out there. I've found out I was doing too much with state/events, and needed to use routes and pages more, and that's taken me to this point.

yoshuawuyts:
Should be fixed in `choo@6.7.0` :tada:

marcbachmann:
With the new choo version, the state in `navigate` now contains also the params and route, so it's equal the state of the new url.
If you need to access the state of the previous url,
you can use `choo.emitter.prependListener('navigate', func)` instead.

PR #559, opened 2017-09-11
Fix toString removing last letter in href

tornqvist:
The regex in `toString` that (what I assume) is supposed to strip out the query from `state.href` also removes the last letter of the href if there's no query.

### DIFF
--- a/index.js
+++ b/index.js
@@ -187,7 +187,7 @@ Choo.prototype.toString = function (location, state) {
   assert.equal(typeof location, 'string', 'choo.toString: location should be type string')
   assert.equal(typeof this.state, 'object', 'choo.toString: state should be type object')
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -187,7 +187,7 @@ Choo.prototype.toString = function (location, state) {
   assert.equal(typeof location, 'string', 'choo.toString: location should be type string')
   assert.equal(typeof this.state, 'object', 'choo.toString: state should be type object')
 
-  this.state.href = location.replace(/\?*.$/, '')
+  this.state.href = location.replace(/\?.+$/, '')
```
jonjaques: image

I believe `/\?.*?$/` is the winner, as the `+` operator modifier will not remove the "?" if that is all that's there.

yoshuawuyts:
`v6.0.1` :tada:

PR #560, opened 2017-09-12
use .storeName during tracing

yoshuawuyts:
This allows tracing individual stores by providing a `.storeName` property on the store. I tried going for `.name` first, but it gets mangled during minification. Ties into https://github.com/choojs/choo-devtools/issues/8; would be cool if we could provide a neato boot cost breakdown, including stores.

The step after this would be to add it to all of our `choojs` org stores.

Thanks!


bcomnes:
Oh nice!

PR #566, opened 2017-09-20
Add link to Reddit.

HoverBaum:
We have an awesome Reddit now, so lets tell people about that.

Check it out: https://www.reddit.com/r/choojs/


bcomnes:
Cool! 👽

PR #568, opened 2017-09-20
chore(pkg): banish lock file

ungoldman:
I noticed in a recent PR there was a commit deleting a generated lock file. I think it would be best disable generation of the lock file for this project to reduce the possibility of lock files being unintentionally committed.


ungoldman:
I also added the two lock files that have become common (yarn & npm) to gitignore.

PR #574, opened 2017-10-04
add initial state support

yoshuawuyts:
Allows checking `window.initialState` when booting, so server rendering actually can do a full rehydration step without any hiccups. Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,9 +47,12 @@ function Choo (opts) {
   this.emitter = nanobus('choo.emit')
 
   var events = { events: this._events }
-  this.state = this.hasWindow && window.initialState
-    ? xtend(window.initialState, events)
-    : events
+  if (this.hasWindow) {
+    this.state = window.initialState
+      ? xtend(window.initialState, events)
+      : events
+    delete window.initialState
+  }
 
   // listen for title changes; available even when calling .toString()
   if (this._hasWindow) this.state.title = document.title
--- a/index.js
+++ b/index.js
@@ -45,7 +45,11 @@ function Choo (opts) {
   // properties that are part of the API
   this.router = nanorouter({ curry: true })
   this.emitter = nanobus('choo.emit')
-  this.state = { events: this._events }
+
+  var events = { events: this._events }
+  this.state = this.hasWindow && window.initialState
+    ? xtend(window.initialState, events)
+    : events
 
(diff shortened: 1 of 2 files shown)


mantoni:
This is also going to help a lot with testing. Very nice 👍

yoshuawuyts:
> Just curious why the initial state is set through global window object? Why not use an argument in choo constructor for instance?

@YerkoPalma I was thinking it'd be neat if it _just worked™_ – I've noticed from talking to people that having a clear cut answer really helps them. I agree that we could also just pass arguments, but this way people need to remember less, and there's no chance of people getting it wrong (or leaking potential security vulnerabilities because we delete the state!)

graforlock:
Is it really a crucial functionality? It adds some code lines for something thats a rather easy thing to implement.

I mean, most of the frameworks leave this to the folks to implement it themselves however they want to hydrate it; since someone might be opinionated etc etc.

yoshuawuyts:
@graforlock it took 4 tries to get this right, haha — in this case we can either add it to Choo core, or document it. I feel that it might lead to a lot of repeated boilerplate across projects if we make it documentation only — and in turn it wouldn't work quite as smoothly with Bankai if it wasn't part of Choo. I feel this is probably the cleanest way forward (:

index.js
```diff
@@ -45,7 +45,16 @@ function Choo (opts) {
   // properties that are part of the API
   this.router = nanorouter({ curry: true })
   this.emitter = nanobus('choo.emit')
-  this.state = { events: this._events }
+
+  var events = { events: this._events }
+  if (this.hasWindow) {
+    this.state = window.initialState
+      ? xtend(window.initialState, events)
```
YerkoPalma: Offtopic. Why are we using `xtend` instead of `Object.assign`?

yoshuawuyts: Browser compat — switching to `Object.assign()` would be a semver major. Agree we should for v7 tho :v:

YerkoPalma:
Is there any reason for this PR to be on hold?

yoshuawuyts:
@YerkoPalma nope, not at all — seems both you and Todd are cool with this, so merging! :sparkles:

yoshuawuyts:
`v6.5.0` :tada:

YerkoPalma:
Awesome! :D :tada:

PR #578, opened 2017-10-08
more timings

yoshuawuyts:
Adds a whole bunch more timings to Choo; allows for better profiling of startup time.

Breaking change b/c `choo.route` and `choo.router` timings are swapped.


tornqvist:
Hmm, looking over this again, I thinking it might be a bit excessive to time `render`, `prerender` and `router`. They're all pretty much just calling each other, so what's really being timed is the timers themselves, right? And that at the cost of creating wrapping functions which themselves add to compute time.

tornqvist:
You mean in Node? That has sorted itself out for me in later versions of Node.

bcomnes:
I just heard something about timings api breaking, but yeah I figured it would be sorted out.  yeah this is fine

PR #584, opened 2017-10-18
Add optionality to 'opts' in Choo constructor

IzumiSy:
This PR makes `opts` parameter in constructor optional, because choo has a code to resolve the parameter if it is undefined: https://github.com/choojs/choo/blob/master/index.js#L20


yoshuawuyts:
📦 `v6.4.3`

PR #585, opened 2017-10-18
Add ES6 'export default' compatibility

IzumiSy:
Currently, this module does not have a good compatibility with ES6 'export default' syntax. This PR adds it.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,7 @@
 
 import * as EventEmitter from 'events'
 
-export = Choo
+export default Choo
 
 declare class Choo {
   constructor (opts: Choo.IChoo)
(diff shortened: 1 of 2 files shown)


IzumiSy:
I know that it is enough to use `import * as choo from 'choo'`, but this PR adds the compatibility with `import choo from 'choo'` style. What do you think?

yoshuawuyts:
Similar reply as in
https://github.com/choojs/choo-log/pull/85#issuecomment-337561552

IzumiSy:
Close this PR with this comment
https://github.com/choojs/choo-log/pull/85#issuecomment-337568198
> I can understand this 👌 I actually would not like to be pushy for this PR. As @tgfjt said at the beginning, it is true that this PR is not always required to those who use ES6 style import syntax. Then, I will close this PR (and this too: choojs/choo#585) for now. Let's continue this discussion again when the right time has come.

PR #586, opened 2017-10-19
Add html/index.d.ts

IzumiSy:
`choo/html` did not have type definition

### DIFF
--- a/html/index.d.ts
+++ b/html/index.d.ts
@@ -1,0 +1,3 @@
+declare function html(strings: TemplateStringsArray, ...keys: any[]): HTMLElement;
+
+export = html
--- a/html/index.d.ts
+++ b/html/index.d.ts
@@ -1,3 +1,2 @@
-declare function html(strings: TemplateStringsArray, ...keys: any[]): HTMLElement;
-
+import * as bel from "bel"
 export = html
--- a/html/index.d.ts
+++ b/html/index.d.ts
@@ -1,2 +1,2 @@
 import * as bel from "bel"
-export = html
+export = bel
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "index.d.ts",
     "html/index.js",
     "html/raw.js",
+    "html/index.d.ts",
     "dist",
     "example"
   ],


tgfjt:
How about requiring [bel](https://github.com/shama/bel/blob/master/bel.d.ts)?

IzumiSy:
Ah, that's also the way.
```
import * as bel from "bel"

export = bel
```
In that case, html/index.d.ts can be like this?

IzumiSy:
I tried that import syntax, but `bel` has a difficulty to import its typing that `bel` does not have `index.d.ts`, but `bel.d.ts` which cannot be imported. This might be supposed to be used by copying it manually to the project typing directory, not by importing it. I don't know why, though :sob: 
(ref: https://github.com/shama/bel/blob/master/bel.d.ts)

 I am going to make a PR to `bel` in order to change the file name of typing file to `index.d.ts` later. In this PR, it would not be bad just to add this small typing file. How about this idea?

tgfjt:
@IzumiSy 
IMO, You can add `"types": "./bel.d.ts"` into bel's `package.json`.  It's compatible.
<https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html>

IzumiSy:
Thank you! I finally made it.

**html/index.d.ts**
```
import * as bel from "bel"
export = bel;
```

**page.js**
```
import html from 'choo/html'

module.exports = function (state: any, emit: any) {
  return html`
    <body>
      <h1>Hello, World!</h1>
    </body>
  `
}
```
Plus, we need a change in bel as you say
> You can add "types": "./bel.d.ts" into bel's package.json.

IzumiSy:
I submitted a change to bel: https://github.com/shama/bel/pull/96

html/index.d.ts
```diff
@@ -0,0 +1,2 @@
+import * as bel from "bel"
+export = html
```
tgfjt: `= html` is correct?

IzumiSy: Ah that's typo.

IzumiSy:
@tgfjt @yoshuawuyts 
https://github.com/shama/bel/pull/96 was successfully merged and now this PR is ready.

yoshuawuyts:
`v6.4.4` :tada:

PR #589, opened 2017-10-22
Fix missing underscore in _hasWindow

tornqvist:
The initial state was not being set to `window.initialState` due to the condition for doing so never resolved to true (the underscore in `this._hasWindow` was missing).


yoshuawuyts:
📦 `v6.5.1`

PR #592, opened 2017-10-26
fix docs

YerkoPalma:
- Use `choo-devtools` in example as `choo-log` got merged into devtools
- Document that stores now accepts the choo instance reference as a third argument


yoshuawuyts:
Neat, thanks!

PR #597, opened 2017-11-06
add support for HTMLElement on choo.mount

moszeed:
Hi, this pull request adds support for HTMLElements, as parameter for choo.mount()

What is done:
- removed explicit "string" test for "selector" variable, replaced by "is available and filled" test
- add "if" to fill "self._tree" by incoming "string" or "HTMLElement" (with is in Browser check)

### DIFF
--- a/index.js
+++ b/index.js
@@ -179,9 +179,7 @@ Choo.prototype.mount = function mount (selector) {
     if (typeof selector === 'string') {
       self._tree = document.querySelector(selector)
     } else {
-      if (selector instanceof window.HTMLElement) {
-        self._tree = selector
-      }
+      self._tree = selector
     }
 
     assert.ok(self._tree, 'choo.mount: could not query selector: ' + selector)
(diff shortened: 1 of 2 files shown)


moszeed:
I`ve done the requested changes.

index.js
```diff
@@ -169,15 +169,21 @@ Choo.prototype.start = function () {
...
 
   var self = this
 
   documentReady(function () {
     var renderTiming = nanotiming('choo.render')
     var newTree = self.start()
+    if (typeof selector === 'string') {
+      self._tree = document.querySelector(selector)
+    } else {
+      if (selector instanceof window.HTMLElement) {
```
yoshuawuyts: In this case, I think it's safe to assume that if it isn't a string, it'll be an HTML element. People would need to try hard to make it fail here I think.

Sorry for being a bit strict here; just out of experience it's that these type of constructions don't compress really well. If file size didn't matter I'd be all for having the stricter check. Thanks!

moszeed: ok :)

Issue #563, opened 2017-09-15
v6.x.x / v7.0.0 - wish list / roadmap


yoshuawuyts:
I want to gather a bit of sentiment for what people might be missing / what you feel might improve. The same as in v6, it'd be nice if all that was needed to update the major was `npm install --save choo@7`, and that's it.

So far my wish list is:
- Figure out how to make routes render async, so we can apply route-based code splitting when needed. Not necessarily breaking, but it might just turn out that we need to rewrite the router (again, sigh - haha).
- Remove the unique labels from the tracing API measures, there's no need for them when using a `PerformanceObserver` like in `choo-log`.
- Replace `xtend` with `Object.assign()` - https://polyfill.io provides a fallback for browsers that don't support it, so for most people this will be a strict improvement.
- `choo/component`, using `nanocomponent` - Ideally it'd be nice if we could formalize tracing / an event system with this too. That way we keep blazing ahead in terms of debugging.
- Reading out from a global state (e.g. `window.initialState`) to provide out of the box "rehydration" of state that was passed on the server.

It's still early, but I'd be interested in hearing input from people on what you find tricky, which things you love, and which things you struggle with! Thanks! :sparkles:

toddself:
I would love if we could figure out a way to get code-splitting to allow for code-split bundles to gain access to `app.use` without using the `window` to store your app instance to it's accessible.

```js
var app = typeof window.app === 'undefined' ? choo : window.app
```

rafaelamorim:
It would be nice to use barracks seamlessly again...
Also, bringing `nanocomponent` closer will be awesome to build some stateful components :)

graforlock:
Point no. 1 - routes rendered async, you might want to check out `universal-router`, i think they got it quite right, I also implemented their router in many side projects of mine.

YerkoPalma:
Some clever way to do page transitions/animations. There has been some discussion before in #487 I'm particulary interested in page transitions, because element component specific animations should be resolved particullary in views or nanocomponents IMO

graforlock:
Extending router would just do, wouldn't it?

`app.route(route, component, beforeRoute)`

tornqvist:
Here's a radical thought regarding async routes: if they were generator functions one could progressively render a view as data/resources become available.

Also, I do miss a standard way of handling errors in choo. Wrapping the route handlers with a  try/catch and emitting a standard error event upon a route crashing or a yielded Promise rejecting would be super!

```javascript
var html = require('choo/html')
[...]

PR #602, opened 2017-11-29
Support async routes

brechtpm:
This probably shouldn't be merged yet, just opening this PR for feedback/discussion.

In this PR I've implemented (part of) the API I proposed [here](https://github.com/choojs/choo/issues/563#issuecomment-344733106). After playing around with it a bit I initially [got stuck](https://github.com/choojs/choo/issues/563#issuecomment-346887984) when trying to make my changes work with the synchronous `toString` API. But then I realized we could just decide not to support `toString` on async routes and throw an AssertionError when anyone tries it. Thoughts?

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,20 +143,19 @@ Choo.prototype.start = function () {
 
   this.emitter.prependListener(self._events.RENDER, nanoraf(function () {
     var renderTiming = nanotiming('choo.render')
-
     self.state.href = self._createLocation()
-    var newTree = self.router(self.state.href)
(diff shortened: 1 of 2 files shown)


brechtpm:
Just FYI, I'm running into some timing issues now using this, so definitely not ready for merge.

brechtpm:
This branch has started to feel a bit too hacky and unstable to me, so unless anyone has some good counterarguments, I think I favour the approach in [this new PR](https://github.com/choojs/choo/pull/603). So gonna close this one for now. If people think this one's better after all we can always reopen.

PR #609, opened 2017-12-13
strip .toString() from browsers

yoshuawuyts:
Was thinking of ways to strip more bytes from the bundle, and `.toString()` was a prime candidate. It's a method that will throw if used in the browser, which makes it quite moot to include in the first place. I can see it happening that we'll include more code for `.toString()` in the future (e.g. async server rendering), most of which would do nothing in the browser.

This PR removes the `.toString()` methods entirely for the browser. Thanks!

## Changes
- split `index.js` into `lib/server.js` and `lib/server.js`
- `index.js` now figures out the right environment, and adds the prototype methods in if needed
- no overhead incurred for browser, as we also added the `browser` property in `package.json`
- also replacing `assert` with `nanoassert` for overhead purposes — should minify better for people not using transforms.


goto-bus-stop:
sweet find :D 

if the `browser` key in package.json points to the browser-only version, does `index.js` still need to check its environment? i think it might be useful to point the `main` key in package.json to the server version, for example if you are testing. you might have a server-side rendering test and a different test that uses jsdom to try to mount the application and test some event handlers or something.

yoshuawuyts:
@goto-bus-stop heya, yeah that's mostly for Electron. Neither the renderer and main processes respect the `browser` key, so the manual check is needed for that unfortunately :(

goto-bus-stop:
Ohh I see, that makes sense then. Thanks for explaining :D



On 15 December 2017 10:28:16 CET, Yoshua Wuyts <notifications@github.com> wrote:
>@goto-bus-stop heya, yeah that's mostly for Electron. Neither the
>renderer and main processes respect the `browser` key, so the manual
>check is needed for that unfortunately :(
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/choojs/choo/pull/609#issuecomment-351957129

-- 
Sent from mobile. Please excuse my brevity.

yoshuawuyts:
ya, actually never mind this. This makes code more complex - and that's one thing we don't need right now.

PR #611, opened 2017-12-19
Allow initial state to be set in options

nullobject:
Allows the inital state to be set in the options passed to the `Choo` function. To set an initial state, just set the `initialState` option.

I also cleaned up the code around setting the initial state from the `window` object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,13 +47,13 @@ function Choo (opts) {
   this.emitter = nanobus('choo.emit')
 
   var events = { events: this._events }
-  if (this._hasWindow) {
-    this.state = window.initialState
-      ? xtend(window.initialState, events)
-      : events
+  var initialState = xtend(opts.initialState, events)
+
+  if (this._hasWindow && window.initialState) {
+    this.state = xtend(window.initialState, initialState)
     delete window.initialState
   } else {
-    this.state = events
+    this.state = initialState
   }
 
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@nullobject what's the use case for this?

heyitsmeuralex:
Surely this is equivalent?

```js
app.use(state => state = {
  // ...
})
```

nullobject:
> what's the use case for this?

@yoshuawuyts The use case is providing a choo application with an initial state object. For example:

```js
var app = choo({initialState: {foo: 'bar'})
```

From the readme:

> When initializing the application, `window.initialState` is used to provision the initial state.

I was initialising my app using `window.initialState`, but I don't think setting a global property on the `window` object is the most obvious way to do this. Passing an option to the `choo` function seems more natural.

nullobject:
> Surely this is equivalent?

@heyitsmeuralex I'm guessing you mean something like [this](https://github.com/choojs/choo/blob/master/example/store.js#L4)? If this is the canonical way to initialise your app state then it should probably be in the readme.

yoshuawuyts:
@nullobject I'm kind of curious what your setup looks like. Are you by any chance running multiple widgets that need initial state?

nullobject:
That's okay, I think I probably misunderstood how to setup my initial state properly.

Issue #610, opened 2017-12-14
Navigate event should enable action on "page entry"


HoverBaum:
### Expected behavior
Listening for the navigate event should enable programmers to do things on "page entry".

### Actual behavior
Navigate event provides no context of from where to where a user is going. `state.route` still holds the old route when the event is fired.

### Use-case

Trigger an event when a user enters a page. This is useful to display messages or fetch data from the backend but only when a user enters a page and not when it is re-rendered.

## Proposal

I could see a few things that could be a solution to this problem and for these use-cases.

1. The Navigate event updated to only trigger listeners once the navigation has happened. As proposed by @yoshuawuyts:
> "navigate triggers" -> "router triggers" -> "state.route is updated" -> "route is triggered" -> "all other listeners for navigate are triggered"
2. The Navigation event carries information about the navigation happening. We could for example add the route we are coming from and going to.

yoshuawuyts:
Should be fixed in `choo@6.7.0` :tada:

Issue #549, opened 2017-08-18
state.query not set until navigate


jondashkyle:
When going to a direct link with a query string, such as `/example?foo=bar`, the `state` object does not contain `query`. But, if you navigate to that URL by clicking a link, it’ll appear as expected.

That seems to be a bug, but just checking the intent before filing a PR. One other question; should `query` always be in `state`, similar to `params`? Could reduce a check or two.

yoshuawuyts:
Yeah, it should always be in there. Definitely sounds like a bug.

mantoni:
Just ran into this as well. The `state.query` object is populated properly, but I have to delay reading from it. It's not yet populated during model initialization.

yoshuawuyts:
Yeah, we should probably document properly that all of these values are
only initialized _after_ DOMContentLoaded

yoshuawuyts:
Should be fixed in `choo@6.7.0` :tada:

PR #613, opened 2017-12-27
Separate route matching from rendering

### LINKED ISSUE #530: State out of date when switching routes

marcbachmann:
This PR
- Fixes the routing to modify the state before the navigate event happens. This means you can access the updated state within the `navigate` event. Modifies the state before the render event is triggered. Therefore this fixes #530, #553, #549, #610
- Exposes `.emit` as an alias to `.emitter.emit`
- Theoretically supports nested routes 😁 . But because we didn't document nor test the router in this repo, I'd handle that as private api for now.

Additionally to the changes I did in here I think we should
  - [x] Define what's the responsiblity of nanorouter & wayfarer and how we could modularize it better
     (I guess this is done by now as nanorouter contains some electron-, and browser specific logic)
  - [x] Deprecate `curry` support in nanorouter. People could just wrap their method by themselves

Please let me know what you think

- [x] Before merging this PR, we have to cleanup the nanorouter dependency.
- [ ] We'll also need to rename some methods and expose the `_handler` differently on the state if we decide to go down this path. In case we want to be able to cancel a navigate event in the future, we could also make it more stateless and pass the new route as object to the navigate events instead of modifying the state.

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,8 +43,9 @@ function Choo (opts) {
   this._tree = null
 
   // properties that are part of the API
-  this.router = nanorouter({ curry: true })
+  this.router = nanorouter()
   this.emitter = nanobus('choo.emit')
+  this.emit = this.emitter.emit.bind(this.emitter)
 
   var events = { events: this._events }
   if (this._hasWindow) {
@@ -68,20 +69,7 @@ function Choo (opts) {
 Choo.prototype.route = function (route, handler) {
   assert.equal(typeof route, 'string', 'choo.route: route should be type string')
   assert.equal(typeof handler, 'function', 'choo.handler: route should be type function')
-
-  var self = this
-  this.router.on(route, function (params) {
-    return function () {
-      self.state.params = params
-      self.state.route = route
-      var routeTiming = nanotiming("choo.route('" + route + "')")
-      var res = handler(self.state, function (eventName, data) {
-        self.emitter.emit.apply(self.emitter, arguments)
-      })
-      routeTiming()
-      return res
-    }
-  })
+  this.router.on(route, handler)
 }
 
 Choo.prototype.use = function (cb) {
@@ -99,7 +87,7 @@ Choo.prototype.start = function () {
   var self = this
   if (this._historyEnabled) {
     this.emitter.prependListener(this._events.NAVIGATE, function () {
-      self.state.query = nanoquery(window.location.search)
+      self._matchRoute()
       if (self._loaded) {
         self.emitter.emit(self._events.RENDER)
         setTimeout(scrollToAnchor.bind(null, window.location.hash), 0)
@@ -136,16 +124,13 @@ Choo.prototype.start = function () {
     }
   }
 
-  this.state.href = this._createLocation()
-  this.state.query = nanoquery(window.location.search)
-  this._tree = this.router(this.state.href)
+  this._matchRoute()
+  this._tree = this._prerender(this.state)
   assert.ok(this._tree, 'choo.start: no valid DOM node returned for location ' + this.state.href)
 
   this.emitter.prependListener(self._events.RENDER, nanoraf(function () {
     var renderTiming = nanotiming('choo.render')
-
-    self.state.href = self._createLocation()
-    var newTree = self.router(self.state.href)
+    var newTree = self._prerender(self.state)
     assert.ok(newTree, 'choo.render: no valid DOM node returned for location ' + self.state.href)
 
     assert.equal(self._tree.nodeName, newTree.nodeName, 'choo.render: The target node <' +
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -202,9 +186,28 @@ Choo.prototype.toString = function (location, state) {
   assert.equal(typeof location, 'string', 'choo.toString: location should be type string')
   assert.equal(typeof this.state, 'object', 'choo.toString: state should be type object')
 
-  this.state.href = location.replace(/\?.+$/, '')
-  this.state.query = nanoquery(location)
-  var html = this.router(location)
+  var href = location.replace(/\?.+$/, '')
+  this._matchRoute(href, location)
```
marcbachmann: We could move this into the `pushState`/`popState`/`replaceState` methods.
But there might be other side effects if we do that.

marcbachmann:
btw. Don't merge just yet 😉

index.js
```diff
@@ -202,9 +186,28 @@ Choo.prototype.toString = function (location, state) {
...
+  this.state.query = nanoquery(queryString)
+  this.state.route = matched.route
+  this.state.params = matched.params
+  this.state._handler = matched.handler
+  return this.state
+}
+
+Choo.prototype._prerender = function (state) {
+  var routeTiming = nanotiming("choo.prerender('" + state.route + "')")
+  var res = state._handler(state, this.emitter.emit.bind(this.emitter))
```
marcbachmann: Should we cache that `this.emitter.emit.bind(this.emitter)` variable somewhere? e.g. in the Choo constructor?

yoshuawuyts: Yeah, that'd be great!

index.js
yoshuawuyts: Perhaps we should leave this the way it is for this PR, and consider the changes in a seperate PR? That might keep the changes nicely contained, and easier to follow along.

index.js
marcbachmann: Directly as `choo.emit`? Then we would have `choo.emitter` and `choo.emit`.

index.js
marcbachmann: Yes, I completely agree here. This definitely needs more elaboration.

marcbachmann:
This is ready now. We could also wait for some time until https://github.com/choojs/nanorouter/pull/13 is merged. In there we've simplified the nanorouter module by removing the curry option.

Let me also know if you want to make the `prerender` and `matchroute` methods private.
e.g.

```diff
@@ -189,7 +189,7 @@ Choo.prototype.toString = function (location, state) {

   var href = location.replace(/\?.+$/, '')
   this._matchRoute(href, location)
-  var html = this._prerender(this.state)
+  var html = prerender(this, this.state)
   assert.ok(html, 'choo.toString: no valid value returned for the route ' + location)
   return html.toString()
 }
@@ -206,9 +206,9 @@ Choo.prototype._matchRoute = function (location, queryString) {
   return this.state
 }

-Choo.prototype._prerender = function (state) {
+function prerender (choo, state) {
   var routeTiming = nanotiming("choo.prerender('" + state.route + "')")
-  var res = state._handler(state, this.emit)
+  var res = state._handler(state, choo.emit)
   routeTiming()
   return res
 }
```

because others like `nanostate` are also exposed, I'm not sure what you prefer.

index.js
```diff
@@ -202,9 +187,28 @@ Choo.prototype.toString = function (location, state) {
...
+  var href = location.replace(/\?.+$/, '')
+  this._matchRoute(href, location)
+  var html = this._prerender(this.state)
   assert.ok(html, 'choo.toString: no valid value returned for the route ' + location)
   return html.toString()
 }
+
+Choo.prototype._matchRoute = function (location, queryString) {
+  if (location === undefined) location = this._createLocation()
+  if (queryString === undefined) queryString = window.location.search
```
marcbachmann: note to myself: I shoud add a comment about that `queryString` is only used in the browser.
Or even move it up to the function call... depending how we'd like to handle it in https://github.com/choojs/choo/pull/609

index.js
marcbachmann: The question here was whether we should prevent `_matchRoute` from mutating the state. It should only return the new params and mutate the state after the `NAVIGATE` event got triggered.
This would allow us to keep the current state but already pass route params to some function to preload something.

This could end up in something similar to the angular `resolve` object where it's possible to preload some resources based on route params. In there we could also cancel the route change if a `resolve` gets rejected.
Maybe this would make everything more complicated. I'll think about it and will create a new issue if I think it's a good idea in the longer term. Here's a link to the resolve reference documentation in case you don't know it: 
https://angular.io/api/router/Resolve

And a few more readable examples:
https://www.codementor.io/max_diachenko/angular-2-route-resolves--typescript-du1089pdm
https://medium.com/@dazcyril/resolving-data-in-angular-2-4-and-5-refactoring-components-and-moving-to-ngrx-store-15e15d8be3dd#9968

johanalkstal: I'm really eager to get this released! :D
Without this it's hard to redirect to other routes based on where the user is trying to go (prevent user from going to routes they should not have access to).

johanalkstal:
I'm really eager to get this released! :D
Without this it's hard to redirect to other routes based on where the user is trying to go (prevent user from going to routes they should not have access to).

marcbachmann:
Nanorouter just got published. So this could land soon in master
...after updating to `nanorouter@3.0.0`

marcbachmann:
@yoshuawuyts this is ready to merge. 
We should tackle some tests soon.

yoshuawuyts:
How should we publish this? Semver... minor, right?

marcbachmann:
It could be a patch as it fixes only the state of the `navigate` event. The sub-route support is just a nice add-on.
Now because the state change happens earlier, this could break something in case somebody accesses some old state value in within the `navigate` event... just as @heyitsmeuralex mentioned.
But because most consumers saw that as bug, I'd do a patch or minor version.

So to sum up, there are the following changes in this PR:
- Fixes the routing to modify the state before the navigate event happens. This means you can access the updated state within the `navigate` event. Modifies the state before the render event is triggered. Therefore this fixes #530, #553, #549, #610
- Exposes `.emit` as an alias to `.emitter.emit`
- Theoretically supports nested routes 😁 . But because we didn't document nor test the router in this repo, I'd handle that as private api for now.

espy:
Sounds like it would also fix https://github.com/choojs/choo/issues/463 😸

yoshuawuyts:
Cool, going to release as minor!

marcbachmann:
❤️ Thanks 🎉

yoshuawuyts:
`choo@6.7.0` :tada:

marcbachmann:
For future reference and people who still want to access the old state within the navigate method,
please use `choo.emitter.prependListener('navigate', func)` instead.

PR #615, opened 2017-12-28
handle multiple arguments for emitter in views

### LINKED ISSUE #614: Emit can't handle more than 1 parameter in views
### Expected behavior

This code (view) should work
```js
function someView (state, emit) {
    return html`
      <body>        
        <p>Send something through sockets:</p>
        <input id="message" />
        <button onclick=${onclick}>Send</button>        
      </body>
    `

    function onclick () {
      emit('ws:send', document.getElementById('message').value, id)      
      document.getElementById('message').value = ''
    }
  }
```

### Actual behavior
The `id` parameter in the `ws:send` event is `undefined`

Put that code in a view

---

This issue is only related to the way choo handle views, it is not related to nanobus. this can be checked running the same example in a store instead of a view, in the store, id is defined. This seems to be due to [this code](https://github.com/choojs/choo/blob/master/index.js#L78), here route handlers (views) are being called with a `eventName` and a single `data` parameter.

**edit**: I guess we could simply change this

```js
      var res = handler(self.state, function (eventName, data) {
        self.emitter.emit(eventName, data)
      })
```

to this

```js
      var res = handler(self.state, self.emitter.emit)
```

YerkoPalma:
fix #614

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,9 +75,7 @@ Choo.prototype.route = function (route, handler) {
       self.state.params = params
       self.state.route = route
       var routeTiming = nanotiming("choo.route('" + route + "')")
-      var res = handler(self.state, function (eventName, data) {
-        self.emitter.emit(eventName, data)
-      })
+      var res = handler(self.state, self.emitter.emit)
       routeTiming()
       return res
     }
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
`v6.6.1` :tada:

PR #625, opened 2018-02-07
server: Store mount selector and return `this`

goto-bus-stop:
This allows `.mount('body')` to be used on the server like so:

```js
module.exports = app.mount('body')
```

Then, when server rendering you can do:

```js
var app = require('./app')
fillServerTemplate(html, app.selector, app.toString('/'))
```

Where `fillServerTemplate` is a function that replaces the HTML at a
CSS selector with some string.

This way you do not need to check for `typeof window` in your
application entry point to do server rendering, and it allows tools like
bankai to automatically inline server rendered html in the correct place.


yoshuawuyts:
Probably also needs a docs update tho

goto-bus-stop:
good call, can you check if the words make sense too?

PR #635, opened 2018-02-23
add support for hyperscript in choo.toString

### LINKED ISSUE #634: add SSR hyperscript support
Hello!

There's been some discussion on IRC about figuring out ways to support hyperscript as a bel alternative when using `choo.toString` in a server context.

## Problem

In a non-browser context, objects returned by hyperscript have a `toString` method that differs from pelo's `toString` method. So currently hyperscript doesn't work in this context.

## Proposal

After doing some investigating, it seems like the easiest way to support hyperscript on the server in choo is to check for `outerHTML` and return that before trying to use the `toString` method returned by [`pelo`](https://github.com/shuhei/pelo).

## Context, Research

[`Element.outerHTML`](https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML) is already a standard feature of the DOM spec that returns the stringified HTML fragment of an element. There's also already a `toString()` method on there that I believe it inherits from [`Object.toString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString), which is the thing that usually returns a very unhelpful string like `[object Object]`.

While my first impulse was to see how hard it would be to modify `hyperscript` to present `choo` with a `toString` method that does the same thing as `pelo().toString()`, it turns out the object hyperscript's factory function returns in a non-browser context is an instance of https://github.com/1N50MN14/html-element, which is already attempting to comply to the standard `Element` spec mentioned before.

## Proposed solution

With all that info in mind, I think it's easiest to modify one line in choo to check for outerHTML first.

Here's the context:

https://github.com/choojs/choo/blob/6bee7086bc0534d46ebad4a6479edf1f6feb207e/index.js#L188-L199

And I think replacing `return html.toString()` with something like this would do the trick:

```js
return html.outerHTML || html.toString()
```

I don't think there's a chance `outerHTML` would be falsey, but we could also do something more defensive like this:

```js
return typeof html.outerHTML === 'string' ? html.outerHTML : html.toString()
```

If this sounds alright to choo maintainers I'll open a PR.

ungoldman:
Resolves #634 (see issue for discussion).

Using the more defensive line to ensure `outerHTML` is a string, otherwise defaulting to `html.toString()`.

I added hyperscript as a dev dependency for the sake of tests, hope that's ok!

### DIFF
--- a/index.js
+++ b/index.js
@@ -195,7 +195,7 @@ Choo.prototype.toString = function (location, state) {
   this._matchRoute(location)
   var html = this._prerender(this.state)
   assert.ok(html, 'choo.toString: no valid value returned for the route ' + location)
-  return html.toString()
+  return typeof html.outerHTML === 'string' ? html.outerHTML : html.toString()
 }
 
 Choo.prototype._matchRoute = function (locationOverride) {
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "bundle-collapser": "^1.2.1",
     "dependency-check": "^2.8.0",
     "discify": "^1.6.0",
+    "hyperscript": "^2.0.2",
     "pretty-bytes-cli": "^2.0.0",
     "spok": "^0.8.1",
     "standard": "^10.0.0",
--- a/test.js
+++ b/test.js
@@ -1,16 +1,28 @@
 var tape = require('tape')
+var h = require('hyperscript')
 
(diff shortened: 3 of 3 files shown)


ungoldman:
@yoshuawuyts your presence is requested ☝️😄

PR #637, opened 2018-03-05
move ._handler off state

yoshuawuyts:
State should be serializable; by moving ._handler off `state` we keep state a bit cleaner. Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -208,17 +208,17 @@ Choo.prototype._matchRoute = function (locationOverride) {
     queryString = window.location.search
   }
   var matched = this.router.match(location)
+  this._handler = matched.cb
   this.state.href = location
   this.state.query = nanoquery(queryString)
   this.state.route = matched.route
   this.state.params = matched.params
-  this.state._handler = matched.cb
   return this.state
 }
 
 Choo.prototype._prerender = function (state) {
   var routeTiming = nanotiming("choo.prerender('" + state.route + "')")
-  var res = state._handler(state, this.emit)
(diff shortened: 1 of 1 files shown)


tornqvist:
Looks good! 👍

goto-bus-stop:
this makes sense to me but I wonder what @marcbachmann thinks because there might've been a reason for doing it this way in #613

marcbachmann:
So far the `state` variable was the only one which was route-specific and mutable, which was the reason I put it there. I'm good with that change. Tests for that would be good as we'll probably run into the same issue again.

yoshuawuyts:
@marcbachmann just to have some context: what issue were you running into? Agree we should have some tests for this; want to make sure we test for the right things.

marcbachmann:
> what issue were you running into?

I was referencing to the statement in your issue description. (State should be serializable).
I had no idea that we wanted to have that state object serializable. How do people use it and do we depend on that?

tornqvist:
> How do people use it and do we depend on that?

I've been experimenting with persisting state to a service worker (for offline support and shared state between tabs) but if you try to post a message with anything but plain serializable values it'll throw.

But yeah, we should probably document the desire to keep state serializable.

yoshuawuyts:
> But yeah, we should probably document the desire to keep state serializable.

yeah, we should probably write a test for that too. I'll do that in a separate patch.

yoshuawuyts:
`v6.10.3`

PR #638, opened 2018-03-06
create stores lazily

yoshuawuyts:
Initialize stores lazily. Opens up the way to run multiple instances of the same app on the server, by passing a custom state all the way down.

This was initially suggested by @diffcunha, and I think it's a pretty decent idea!

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,6 +40,7 @@ function Choo (opts) {
   this._hasWindow = typeof window !== 'undefined'
   this._createLocation = nanolocation
   this._loaded = false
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
i like it!! we can probably also build on this to fix the issue with async rendering on the server that @tornqvist mentioned in https://github.com/choojs/bankai/pull/427#issuecomment-368890097

yoshuawuyts:
@goto-bus-stop yes, exactly!

PR #639, opened 2018-03-07
choo components

yoshuawuyts:
Replacement for #606. Thanks!


index.js
```diff
@@ -235,3 +243,23 @@ Choo.prototype._prerender = function (state) {
   routeTiming()
   return res
 }
+
+Choo.prototype._setCache = function (state) {
+  var cache = new Cache(state, this.emitter.emit.bind(this.emitter), this._cache)
+  state.cache = renderComponent
+
+  function renderComponent (Component, id) {
+    assert.equal(typeof Component, 'function', 'choo-component-preview: Component should be type function')
```
YerkoPalma: I guess this is no longer a `preview` right?

component/cache.js
```diff
@@ -0,0 +1,40 @@
...
+
+module.exports = ChooComponentCache
+
+function ChooComponentCache (state, emit, lru) {
+  assert.ok(this instanceof ChooComponentCache, 'ChooComponentCache should be created with `new`')
+
+  assert.equal(typeof state, 'object', 'ChooComponentCache: state should be type object')
+  assert.equal(typeof emit, 'function', 'ChooComponentCache: state should be type function')
+
+  this.cache = lru || new LRU(100)
```
goto-bus-stop: An LRU cache would start evicting components that are still on the page if you have a lot of them (eg. 120). I think components that are still mounted should always stay in the cache. Maybe we can have two caches, one plain object `.mounted`, and one LRU (or user provided) `.unmounted`. Then we can use on-load to move components into the unlimited `.mounted` cache on `load`, and move them to the bounded `.unmounted` cache on `unload`.

example/components/footer/index.js
```diff
@@ -0,0 +1,54 @@
+var Component = require('../../../component')
+var html = require('bel')
+
+var clearButton = require('./clear-button')
+var filterButton = require('./filter-button')
+
+module.exports = class Footer extends Component {
+  static identity () {
+    return 'footer'
+  }
```
goto-bus-stop: the `identity` methods are no longer used, right?

component/cache.js
s3ththompson: simpler is probably better, and i like your suggestion @goto-bus-stop, but here's one other idea: i wonder if it makes sense to have separate caches per component class? i guess it would just be nice not to thrash header/footer/button singleton components if an app happens to have like 1000 tweet component instances...

goto-bus-stop: That would help too! Id still be worried about the tweets though. as I understand it a list of 100+ tweets would end up recreating and evicting instances during every render even if their content didn't change...

goto-bus-stop: The onload idea might be pretty difficult to implement in a generic way, though, I'm not sure.

yoshuawuyts: > Id still be worried about the tweets though

If people are thinking of creating lots and lots of elements, we could recommend they create a separate cache for this specifically. If we expose the cache instance, they'll have the flexibility of choosing elements and such.

yoshuawuyts:
Some more eyes on this would be great btw! @choojs/trainspotters

ungoldman:
Not to derail but would this make more sense as a middleware rather than shipping with core?

jondashkyle:
@ungoldman First-class support for components in the framework helps tremendously with the accessibility effort. Having both `state` and `emit` exposed without any prop juggling in userland being most notable gain, personally.

yoshuawuyts:
@ungoldman yep, what @jondashkyle says. Components have shown to be a requirement people have in almost every project. Being able to use them straight from core lowers the barrier, and allows us to assume people can build on top of it in every project.

YerkoPalma:
I'm ok with this, but I think docs should also be updated, to reflect the use of components.

yoshuawuyts:
ping @tornqvist @bcomnes @toddself your input on this patch would be greatly appreciated :pray:

index.js
```diff
@@ -39,25 +41,29 @@ function Choo (opts) {
...
   this._createLocation = nanolocation
+  this._cache = opts.cache
   this._loaded = false
   this._stores = []
   this._tree = null
 
-  // properties that are part of the API
-  this.router = nanorouter()
-  this.emitter = nanobus('choo.emit')
-  this.emit = this.emitter.emit.bind(this.emitter)
```
tornqvist: Removing `this.emit` breaks `_prerender`. Making sure that emit is a reference to the same function on every render (by referring a bound method on self) really helps with shallow diff of arguments. Also saves memory not having to allocate a new function on every render.

component/cache.js
tornqvist: +1 on what @yoshuawuyts said but to make that easier couldn't we allow for the third argument (`lru`) to be a number and forward that to `nanocache`? So if you don't have any preference as to what lru cache to use, but just want to allow for a lot of instances you'd do `var manyTweetsCache = new ComponentCache(state, emit, 250)`

Also, most classes in the choo stack allows you to create instances without `new` using an `instanceof` check and calling `new` for you if needed. To maintain a uniform api, shouldn't we do that here as well?

index.js
yoshuawuyts: Oopts, that was totally accidental. Great catch!

yoshuawuyts: Fixed & added regression tests :sparkles:

component/cache.js
```diff
@@ -0,0 +1,40 @@
...
+// Get & create component instances.
+ChooComponentCache.prototype.render = function (Component, id) {
+  assert.equal(typeof Component, 'function', 'ChooComponentCache.render: Component should be type function')
+  assert.ok(typeof id === 'string' || typeof id === 'number', 'ChooComponentCache.render: id should be type string or type number')
+
+  var el = this.cache.get(id)
+  if (!el) {
+    var args = []
+    for (var i = 0, len = arguments.length; i < len; i++) {
+      args.push(arguments[i])
```
tornqvist: We should probably drop `Component` and `id` from the args here (i.e. start iterating on `2`). They are both prefixed on the next line anyway and by doing this we'd be offsetting the actual constructor arguments by an additional two.
This is what we're doing now:
```javascript
class Foo extends Component {
  constructor (id, state, emit, the_class, id, actual_constructor_argument) { … }
}
```

yoshuawuyts: Nice, done!

yoshuawuyts:
Merged in @tornqvist's feedback. Thinking of merging this in 24 hours, and releasing it as a pre-release. Then follow up with a proper release next week. I think it's time to move this forward.

Please do chime in if there's any problems left with the implementation. If not, we'll be able able to try it out from a pre-release soon! :grin:

component/cache.js
```diff
@@ -0,0 +1,41 @@
+var assert = require('assert')
+var LRU = require('nanolru')
+
+module.exports = ChooComponentCache
+
+function ChooComponentCache (state, emit, lru) {
+  assert.ok(this instanceof ChooComponentCache, 'ChooComponentCache should be created with `new`')
+
+  assert.equal(typeof state, 'object', 'ChooComponentCache: state should be type object')
+  assert.equal(typeof emit, 'function', 'ChooComponentCache: state should be type function')
```
bendik: *emit* should be type function 🎹

yoshuawuyts: thanks!

Flet:
🎉 🎉 🎉 🎉 🎉

yoshuawuyts:
`v6.11.0-preview1`

PR #640, opened 2018-03-08
show total downloads

yoshuawuyts:
Show total downloads, rather than downloads / month. Feel it removes a bit of the pressure to amp up monthly downloads - instead emphasizing how we've been doing over the past few years :sparkles:


kareniel:
108K  :D

PR #641, opened 2018-03-08
Add gitignore inside example/

goto-bus-stop:
`"example"` is in the `"files"` key in our package.json, which overrides
the root .gitignore and ends up including the entire examples folder,
including its dist/ and node_modules/. That means a fresh install of
choo comes in at over 100MB of unused code.

Nested .gitignores still apply to folders that are specified in the
`files` key, so this pushes choo back down to a <500KB installed size :)
The example code itself is still included so folks can use it as
offline documentation.

### DIFF
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -1,2 +1,3 @@
 dist/
 node_modules/
+package-lock.json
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
thought this was a pretty big deal so i merged it w/o approvals—feel free to revert if anyone prefers a different solution!

PR #642, opened 2018-03-12
assert for arrays on .toString()

yoshuawuyts:
Found out about an edge case through DMs, reroduced in https://github.com/yoshuawuyts/repro-nanohtml-render-array. Apparently we were allowing top-level arrays to slip through, causing all sorts of funky rendering. This fixes that. Thanks!


yoshuawuyts:
`v6.10.2` :tada:

PR #646, opened 2018-03-14
Asynchronous render

### LINKED ISSUE #645: Asynchronous render
Rendering in Choo is a completely synchronous process, thought this makes things way simpler there are a few downsides from 2 perspectives:

- **CPU** - Rendering complex components might block the main loop affecting the overall experience 
- **I/O** - It's not possible to make the render wait for asynchronous I/O events such as data loading

I believe the later is the most relevant when it comes to SSR. One of the biggest issues of sync SSR is related to data loading and code-splitting. There are several workaround strategies to deal with async I/O on a sync rendering process, most of them require a first-pass render to load or collect something you can wait on and signal when the data is available to read synchronously. Besides being a waste of CPU cycles it's also not very elegant and adds unnecessary complexity to the code. One could just not have SSR but being able to have truly isomorphic or universal app is a requirement for most content-driven apps.

When I think of async rendering I think of composing the view using sync or async components and then one of two things can happen:

- Wait until all the components are resolved and then apply node diffing
- Apply node diffing as components resolve

Ideally both scenarios should be allowed, giving that choice to the developer, but I believe the first one is easier to start with given the current Choo architecture.

In order to achieve the the first scenario this would have to be allowed:

```js

### Actual behavior
Write here.

Write here.

diffcunha:
Fixes #645


toddself:
This code requires that any browser or node instance that doesn't support async be compiled using babel and the polyfill being included.

Being that we still have support enabled for IE 11 and Safari (browser which represent more than 1% of usage), this adds a lot of weight to choo. (babel-poylfill is around 85k depending on compression which is MUCH larger than choo)

diffcunha:
I just realised that, I'm removing async/await now

toddself:
Also, check out my comments in #645 -- there are other ways to do this without needing to change how choo works

diffcunha:
I forgot to mention but this change is also backward compatible, normal sync rendering will still work

toddself:
`Promise` still needs to be shimmed for IE 11 (so we'd likely also need to include a promises implementation in core).

diffcunha:
Does it have to be included in core? It could be up to the developer to decide if she wants it or not and  which shim to use. Otherwise we could use `es6-promise` in core, it takes 2.4 KB gziped

YerkoPalma:
I think this whole PR could fit for a store. Since the the `.use` function expose the app instance, you can change whatever you want there, maybe is a good idea to have it as astore and make some tests and bench on top of it.

diffcunha:
@YerkoPalma that's actually a good idea. I could definitely do it but, there is a recent change to delegate store initialisation (https://github.com/choojs/choo/pull/638), it makes it harder since the patching would only be applied after calling `start` or `toString`

diffcunha:
@toddself I updated the PR so that the a `Promise` shim is only required if the developer wants to go with async rendering. If there are no promises being passed to `html`, everything works just like before.

mcollina:
I really like this implementation, as it brings the best of both worlds and it's not a breaking change.
I would add some more tests with mixed values (promises, literals) as well.

Powersource:
Any update on this? :)

Anything a noob contributor (me lol) could help out with?

PR #653, opened 2018-04-04
WIP Experiment Async Route support

yoshuawuyts:
Porting https://github.com/goto-bus-stop/choo-lazy-route to be a first-class citizen in Choo. Should allow loading routes async directly.

Right now the API is prefixed with experimental, as it probably requires some discussion on how to stabilize. But feel it's worthwhile adding a non-hacky way to allow this in the meantime. If anything so we can iterate on the functionality while keeping clear that this isn't ready for prime time yet.

Thanks!


index.js
```diff
@@ -76,10 +78,66 @@ function Choo (opts) {
...
+
+  this.route(route, function (state, emit) {
+    if (!self._asyncProxy) self._initAsyncProxy()
+    // Begin loading the bundle on the first call
+    if (loadingState === IDLE) {
+      emit('choo:async-route-start', state.route)
+      renderRoute = state.route
+      loadingState = LOADING
+
+      var p = loader(onload)
```
goto-bus-stop: we don't need to pass in the `onload` callback here if `p` must return a promise anyway

tornqvist:
I love this idea! Before I knew about choo-lazy-route I started work on pretty much the exact same thing called [choo-lazy-view](https://github.com/jallajs/choo-lazy-view) but it has support for a third argument being a loader view. Not only was the loader view a nice way of giving immediate feedback but also offers an opportunity for prefetching content needed by the view being loaded. Ofc, this could just as well be done by listening for the `choo:async-route-start`. It also defaults to returning the mounted node as loading view to prevent flash of an empty view.

yoshuawuyts:
@tornqvist it'd be awesome if someone could take this PR over the finish line. I ran out of time to work on it in over summer, but would love if this could eventually land!

tornqvist:
I've been working a bit on this lately but have hit a wall regarding what to do with SSR. On the client side we either preserve the current view as the async route is being loaded or the user could supply a loader function that would be rendered in its place.

On the server side it's a bit more tricky. Both [choo-lazy-route](https://github.com/goto-bus-stop/choo-lazy-route) and [choo-lazy-view](https://github.com/jallajs/choo-lazy-view) use `state.prefetch` (or `state._experimental_prefetch`) to inform the user when the view has loaded, encouraging a second render pass.

As I see it we either standardize `state.prefetch` as a core choo pattern for these kinds of things or we allow `toString` ro return a promise, in which case the name doesn't really make sense and would need change to e.g. `render`.

Any opinions on this matter? I personally would prefer the latter and let `state.prefetch` just be an implementation detail, which could easily be used in conjunction with an async `render` method.

PR #667, opened 2018-06-18
Hash routing opt in

### LINKED ISSUE #440: handling both hash routing & anchor links
As I'm porting [minidocs to choo v5](https://github.com/freeman-lab/minidocs/tree/choo-v5) I'm working on making choo v5 support both hash routing & navigating to anchor links if they are present on the page.

The tricky part is that when the router is first called, the element isn't available on the page yet: https://github.com/yoshuawuyts/choo/blob/master/index.js#L50

So there's not a way to check if an id is available to scroll to before the router is called.

Thoughts on how to implement this?

### LINKED ISSUE #666: Page anchors ignored
### Expected behavior
When an anchor tag with a hash route is clicked, `<a href="#page-anchor">Jump To</a>`, the page should first be checked to see if an available page anchor exists and update scroll. Only if a page anchor is not found should a route change be triggered. Basing this off of https://github.com/choojs/choo#hash-routing and https://choo.io/docs/routing

### Actual behavior
Route change is always triggered and no jump-to behavior is possible

### Example
demo: https://choo-hash-bug.glitch.me/
code: https://glitch.com/edit/#!/choo-hash-bug?path=index.js:17:23

Curious if anyone else has ran into this? ✌️

tornqvist:
Resolves #666, resolves #440 

Both newcomers as well as seasoned choo users keep stumbling over this and resolve to hackish workarounds or complex subclassing of choo.

Making hash routing *opt-in*, hash anchors will behave as expected as per browser defaults. By opting into hash routing hashes will be transformed to be a part of the pathname (this is the current default behavior).

Example with default new behavior: https://choo-hash-anchors.glitch.me
Example with hash routing enabled: https://choo-hash-anchors-enabled.glitch.me/my/app

I ended up factoring out [nanolocation](https://github.com/choojs/nanolocation) as the behavior had to be duplicated for location override.

This is most definitely a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ containing the key-value mappings exists as `state.query`.
 ### Hash routing
 By default hashes are not taken into account when routing. Using hashes to
 delimit routes (e.g. `/foo#bar`) is supported if the `hash` options set to
-`true`. Regardles, when a hash is found we also check if there's an
+`true`. Regardless, when a hash is found we also check if there's an
 available anchor on the same page, and will scroll the screen to the position.
 Using both hashes in URLs and anchor links on the page is generally not
 recommended.
--- a/README.md
+++ b/README.md
@@ -284,12 +284,12 @@ Querystrings (e.g. `?foo=bar`) are ignored when matching routes. An object
 containing the key-value mappings exists as `state.query`.
 
 ### Hash routing
-By default hashes are not taken into account when routing. Using hashes to
-delimit routes (e.g. `/foo#bar`) is supported if the `hash` options set to
-`true`. Regardless, when a hash is found we also check if there's an
-available anchor on the same page, and will scroll the screen to the position.
-Using both hashes in URLs and anchor links on the page is generally not
-recommended.
+By default hashes are treated as part of the url when routing. Using hashes to
+delimit routes (e.g. `/foo#bar`) can be disabled by setting the `hash`
+[option](#app--chooopts) to `false`. Regardless, when a hash is found we also
+check if there's an available anchor on the same page, and will scroll the
+screen to the position. Using both hashes in URLs and anchor links on the page
+is generally not recommended.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -276,10 +276,12 @@ Querystrings (e.g. `?foo=bar`) are ignored when matching routes. An object
 containing the key-value mappings exists as `state.query`.
 
 ### Hash routing
-Using hashes to delimit routes is supported out of the box (e.g. `/foo#bar`).
-When a hash is found we also check if there's an available anchor on the same
-page, and will scroll the screen to the position. Using both hashes in URLs and
-anchor links on the page is generally not recommended.
+By default hashes are not taken into account when routing. Using hashes to
+delimit routes (e.g. `/foo#bar`) is supported if the `hash` options set to
+`true`. Regardles, when a hash is found we also check if there's an
```
toddself: missing the last `s` in regardless.

toddself:
This is a great idea.  It'll need to be a semver major right?

toddself:
Oops that dang close button is so close to the comment button :)

YerkoPalma:
I'm not clear why this would be a major change. In what case would it be breaking stuff?

tornqvist:
Since hash routing is enabled by default, changing it to be opt-in would break any app that currently relies on hashes being interpreted as part of the pathname.

goto-bus-stop:
How about adding the hash option with default true in a minor release and switching to false in v7?

tornqvist:
A minor version defaulting to true would be a great way of getting this out there asap. I’m currently on vacation but can resolve the conflict and change the default in a day or two.

tornqvist:
I switched it around to default hash routing to `true` so that we can release this as a minor. Merging this now.

tornqvist:
v6.13.0 🎉

PR #672, opened 2018-07-03
Render cached components by id

### LINKED ISSUE #670: Rendering cached components should be done by id only
### Expected behavior
After rendering a component like 

```js
state.cache(PostPreview, 'post-id').render()
```

next uses of the same component should be done from cache like

```js
state.cache('post-id').render()
```

### Actual behavior
This call

```js
state.cache('post-id').render()
```

throws an exception. Even this call

```js
state.cache(null, 'post-id').render()
```

throws an exception. the only way to do this is by passing a noop to the cache method.

YerkoPalma:
fix #670

### DIFF
--- a/component/cache.js
+++ b/component/cache.js
@@ -17,6 +17,10 @@ function ChooComponentCache (state, emit, lru) {
 
 // Get & create component instances.
 ChooComponentCache.prototype.render = function (Component, id) {
+  if (!id) {
+    id = Component
+    Component = function () {}
+  }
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
I think having to use the `state.cache(Component, 'id')` API everywhere is intentional. There are no guarantees for anything actually being in the cache, so `state.cache('id').render()` could start throwing if something was removed.

YerkoPalma:
Oh I see, closing then.

Thanks for the response.

PR #673, opened 2018-07-07
Add documentation on components

### LINKED ISSUE #662: Document components
### Expected behavior
Since component support has landed in Choo, I should know how to use it without reading the code.

### Actual behavior
There is no doc about components.

tornqvist:
Resolves #662 

A first take on documentation of components in Choo and the new component cache. I wasn't sure about how much detail to go into about the component API seeing as nanocomponent has some excellent docs of its own. Also, I'm not 100% sure it's the best idea to drag in mapbox and and the geolocation API in the example but seeing as the text speaks of 3rd party libraries, it made sense to illustrate that. Maybe the nanocomponent [button example](https://github.com/choojs/nanocomponent#usage) or something like that would suffice?

Would very much appreciate feedback from @yoshuawuyts, @bcomnes and @s3ththompson as they were instrumental in brining this into Choo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@
 - [State](#state)
 - [Routing](#routing)
 - [Server Rendering](#server-rendering)
+- [Components](#components)
 - [Optimizations](#optimizations)
 - [FAQ](#faq)
 - [API](#api)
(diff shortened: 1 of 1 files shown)


YerkoPalma:
LGTM!

PR #528, opened 2017-07-03
Add v6 tests (WIP)

timwis:
Not ready to merge yet, work in progress. Jotted down ideas for tests below. Let me know if any don't make sense, and feel free to add other ideas.

## Unit test ideas
- [x] enables history by default
- [ ] if history enabled, registers navigate, popstate, pushstate, replacestate events
- [x] enables href by default
- [ ] if href enabled, clicking link trigers pushstate (failing, likely due to nanohref cancelling in node)
  - [ ] unless link is current location
- [ ] `.route` registers the route
- [x] route is passed state and emit fn
  - [ ] verify emit fn context?
- [x] state is passed current route
- [ ] state is passed current search query (jsdom doesn't support changing search)
- [x] callback passed to `.use` is passed state, emitter, and app
  - [ ] verify context?
- [ ] on navigate event, emits render
- [ ] on navigate event, updates query
- [ ] on navigate event, scrolls to anchor
- [ ] on popstate event, emits navigate
- [ ] on pushstate event, modifies history and emits navigate
- [ ] on replacestate event, replaces history and emits navigate
- [ ] on window.onpopstate, emits popstate event
- [ ] domcontentloaded event fired on document ready
- [ ] mount morphs the target node
- [ ] toString works
- [ ] on render event, tree is morphed

## higher level (closer to integration tests)
- [ ] state change re-renders? (ie. choo 101 app)
- [ ] `*` route called by default
  - [ ] `/*` route works too (maybe nanorouter)

PR #674, opened 2018-07-15
Add v6 tests

timwis:
Replaces #528. I was just going off the features listed in the readme. There are a handful more we should test but this is a start.

Note that this includes a readme change for `state.href` that I believe was incorrectly documented. Let me know if I got that wrong.

Also note that we should delete the original `v6-tests` branch once this is merged. Leaving it for now for reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ An object containing the current queryString. `/foo?bin=baz` becomes `{ bin:
 'baz' }`.
 
 ### `state.href`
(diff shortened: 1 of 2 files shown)


tornqvist:
I've added tests for the new component cache. Would it be alright if I push them to this branch?

bcomnes:
Up to Tim, but I'm cool with it.  We should land these.

timwis:
Absolutely!

bcomnes:
🎉

PR #689, opened 2018-10-30
Fix wrong this usage in nanohref integration

mantoni:
I currently see `TypeError: undefined is not an object (evaluating 'this._hashEnabled')` being thrown by clients. This should fix it.


mantoni:
Impressive turnaound time. Thanks a lot!

goto-bus-stop:
:package: 6.13.1

PR #691, opened 2018-11-14
typings: add IChoo.hash

seangenabe:


toddself:
We should add a `cache` option too.

Would you mind adding it?

Looks like the definition would be:

`cache: number | {max:number, maxAge:number}`

toddself:
Also thanks for the PR @seangenabe

PR #695, opened 2019-04-01
Fix location missing on store init

tornqvist:
Location meta data was missing on store init during SSR, meaning `href`, `query`, `route` and `params` was missing from state when initializing stores.

This fixes that by running `_matchRoute` before initializing stores during SSR.


tornqvist:
v6.13.2 🎉

PR #697, opened 2019-05-28
Update README.md

perguth:
Add missing div ID so Mapbox can mount.


perguth:
Not sure about this though. Why would the ID the be configurable on the components interface?

perguth:
I misread that line:

> The component cache is a function which takes a component class **and a unique id (string)** as it's first two arguments.

PR #698, opened 2019-06-08
Match route before init stores

tornqvist:
This adds back the changes made in #695 ensuring that `href`, `query`, `params` etc. are in the state _before_ the stores are initialized.

### DIFF
--- a/index.js
+++ b/index.js
@@ -138,11 +138,11 @@ Choo.prototype.start = function () {
   }
 
   this._setCache(this.state)
+  this._matchRoute()
   this._stores.forEach(function (initStore) {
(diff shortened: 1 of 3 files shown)


bcomnes:
Is this a breaking change?

tornqvist:
It could be, if a stores is checking for these properties in `state` to determine if it has been initialized before or not. Also, in combination with https://github.com/choojs/choo/pull/649 these properties will always be in state on initialization, as opposed to only after the first call `toString`.

@yoshuawuyts mentioned on IRC that the init order has been switched around before, but not as a breaking change and apparently it caused issues for some people.

PR #718, opened 2026-02-19
fix: correct typo 'overriden' → 'overridden' in README

slegarraga:
Small typo fix: `overriden` → `overridden` (double 'd').


goto-bus-stop (approved):
thanks!

